PR #14, opened 2016-05-18
Set to field when using merge_data

lewistaylor:
The `to` field is required even if providing recipient addresses in x-smtpapi. See https://sendgrid.com/docs/API_Reference/Web_API/mail.html#-send.

### DIFF
--- a/anymail/backends/sendgrid.py
+++ b/anymail/backends/sendgrid.py
@@ -100,6 +100,7 @@ class SendGridPayload(RequestsPayload):
             self.set_recipients('to', self.to_list)
         else:
             # Merge-friendly smtpapi 'to' field
+            self.set_recipients('to', self.to_list)
             self.smtpapi['to'] = [email.address for email in self.to_list]
             self.all_recipients += self.to_list
 
--- a/tests/test_sendgrid_backend.py
+++ b/tests/test_sendgrid_backend.py
@@ -432,7 +432,7 @@ class SendGridBackendAnymailFeatureTests(SendGridBackendMockAPITestCase):
         data = self.get_api_call_data()
         smtpapi = self.get_smtpapi()
         self.assertEqual(data['toname'], [' ', 'Bob'])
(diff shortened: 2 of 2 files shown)


lewistaylor:
Please note I'm stacked with another project and haven't added any regression tests for it.

medmunds:
Thanks for this.

I'm a little confused by [SendGrid's docs](https://sendgrid.com/docs/API_Reference/Web_API/mail.html#-send) on this. As you point out, `to` is required and "Must be a valid email address." But:

> Note that recipients passed in this parameter will be visible as part of the message. If you wish to hide the recipients, use the TO parameter in the x-smtpapi header.

With merge_data (batch send), we definitely want to hide the recipients from each other. If you provide both `to` and `x-smtpapi to`, do all the recipients see the full `to` list?

medmunds:
>  If you provide both `to` and `x-smtpapi to`, do all the recipients see the full to list?

OK, quick test answered my own question: with both `to` params set, SendGrid does *not* make all the recipients visible to each other.

Thanks for the fix. I'll add an integration test for this, too.

avelis:
@medmunds I had a chance to talk to SendGrid support on this awhile back. x-smtpapi `to` field will not show all the recipients emails to each other. The regular `to` field will.

When using the x-smtpapi `to` field the regular `to` field will be ignored. Therefore they suggested adding the `from` email as the `to` in this specific scenario.

I do agree that the SendGrid docs don't make this clearer.

lewistaylor:
@avelis @medmunds I'll update the code and tests later today

medmunds:
@avelis thanks- that's helpful to know. Using `from` as `to` is probably a good safety mechanism to avoid accidentally exposing all the real `to` emails if something went wrong with smtpapi.

@lewistaylor I already released your PR (with a couple of tweaks) as 0.3.1. You may need to rebase before updating this PR. (Or just submit a new one if that's easier.)

PR #25, opened 2016-06-24
fix project url on setup.py

luzfcb:


medmunds:
Nice catch. Thanks!

luzfcb:
:+1:

PR #42, opened 2016-10-31
Compatibility with Pythons older than 2.7.7

### LINKED ISSUE #41: Webhooks in mailgun backend only work with newest 2.7 pythons
`hmac.compare_digest` is [new in 2.7.7](https://docs.python.org/2/library/hmac.html#hmac.compare_digest). Debian Wheezy, LTS supported until May 2018, is [currently at 2.7.3](https://packages.debian.org/search?keywords=python2.7&searchon=names&exact=1&suite=all&section=all).

As a consequence, trying to use Mailgun mailhooks on Wheezy fails.

I suppose the fix is [something like this](https://github.com/liris/websocket-client/commit/4d5bc1e4e3622265ef7dc54281c0f73bb4dd98d6), along perhaps with a deprecation notice or similar?

sebbacon:
Fixes #41

### DIFF
--- a/anymail/webhooks/mailgun.py
+++ b/anymail/webhooks/mailgun.py
@@ -3,20 +3,13 @@ from datetime import datetime
 
 import hashlib
 import hmac
+from django.utils.crypto import constant_time_compare
 from django.utils.timezone import utc
 
 from .base import AnymailBaseWebhookView
 from ..exceptions import AnymailWebhookValidationFailure
 from ..signals import tracking, AnymailTrackingEvent, EventType, RejectReason
 from ..utils import get_anymail_setting, combine
-
-# Use safer comparison method if available
-# https://docs.python.org/2/library/hmac.html#hmac.compare_digest
-if hasattr(hmac, "compare_digest"):
-    compare_digest = hmac.compare_digest
-else:
-    def compare_digest(s1, s2):
-        return s1 == s2
 
 
 class MailgunBaseWebhookView(AnymailBaseWebhookView):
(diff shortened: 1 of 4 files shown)


anymail/webhooks/mailgun.py
```diff
@@ -10,6 +10,14 @@
...
 from ..utils import get_anymail_setting, combine
 
+# Use safer comparison method if available
+# https://docs.python.org/2/library/hmac.html#hmac.compare_digest
+if hasattr(hmac, "compare_digest"):
+    compare_digest = hmac.compare_digest
+else:
+    def compare_digest(s1, s2):
+        return s1 == s2
+
```
medmunds: I'd suggest instead switching to [`django.utils.crypto.constant_time_compare`](https://github.com/django/django/blob/stable/1.8.x/django/utils/crypto.py#L79-L84), which is available on all Django versions we support. That preserves the security aspect of the code (unlike a simple `==` compare).

anymail/webhooks/mailgun.py
```diff
@@ -34,7 +42,7 @@ def validate_request(self, request):
             raise AnymailWebhookValidationFailure("Mailgun webhook called without required security fields")
         expected_signature = hmac.new(key=self.api_key, msg='{}{}'.format(timestamp, token).encode('ascii'),
                                       digestmod=hashlib.sha256).hexdigest()
-        if not hmac.compare_digest(signature, expected_signature):
+        if not compare_digest(signature, expected_signature):
```
medmunds: The same change is also needed in the Mandrill webhook.

PR #53, opened 2017-03-02
WIP Mailjet backend

calvin:
Still very early in the stage, but I opened this PR to get some feedbacks.

Todos:

- [ ] add mock server
- [ ] add tests

### DIFF
--- a/anymail/backends/mailjet.py
+++ b/anymail/backends/mailjet.py
@@ -1,0 +1,124 @@
+from ..exceptions import AnymailRequestsAPIError
+from ..message import AnymailRecipientStatus, ANYMAIL_STATUSES
+from ..utils import get_anymail_setting
+
+from .base_requests import AnymailRequestsBackend, RequestsPayload
+
+
+class EmailBackend(AnymailRequestsBackend):
+    """
+    Mailjet API Email Backend
+    """
+
+    esp_name = "Mailjet"
+
+    def __init__(self, **kwargs):
+        """Init options from Django settings"""
+        esp_name = self.esp_name
+        self.api_key = get_anymail_setting('api_key', esp_name=esp_name, kwargs=kwargs, allow_bare=True)
+        api_url = get_anymail_setting('api_url', esp_name=esp_name, kwargs=kwargs,
+                                      default="https://api.mailjet.com/v3")
+        if not api_url.endswith("/"):
+            api_url += "/"
+        super(EmailBackend, self).__init__(api_url, **kwargs)
+
+    def build_message_payload(self, message, defaults):
+        return MailjetPayload(message, defaults, self)
+
+    def parse_recipient_status(self, response, payload, message):
+        parsed_response = self.deserialize_json_response(response, payload, message)
+        recipient_status = {}
+        try:
+            for key in parsed_response:
+                status = key.lower()
+                if status not in ANYMAIL_STATUSES:
+                    status = 'unknown'
+
+                for item in parsed_response[key]:
+                    message_id = item.get('MessageID_id', None)
+                    email = item['Email']
+                    recipient_status[email] = AnymailRecipientStatus(message_id=message_id, status=status)
+        except (KeyError, TypeError):
+            raise AnymailRequestsAPIError("Invalid Mailjet API response format",
+                                          email_message=message, payload=payload, response=response,
+                                          backend=self)
+        return recipient_status
+
+
+class MailjetPayload(RequestsPayload):
+
+    def __init__(self, *args, **kwargs):
+        self.esp_extra = {}  # late-bound in serialize_data
+        self.headers = {
+            'Content-Type': 'application/json',
+        }
+        super(MailjetPayload, self).__init__(*args, **kwargs)
+
+    def get_api_endpoint(self):
+        return "messages/send"
+
+    def serialize_data(self):
+        return self.serialize_json(self.data)
+
+    #
+    # Payload construction
+    #
+
+    def init_payload(self):
+        self.data = {
+        }
+
+    def set_from_email(self, email):
+        self.data["FromEmail"] = email.email
+        if email.name:
+            self.data["FromName"] = email.name
+
+    def add_recipient(self, recipient_type, email):
+        assert recipient_type in ["to", "cc", "bcc"]
+        if email.name:
+            address = "{} <{}>".format(email.name, email.email)
+        else:
+            address = "<{}>".format(email.email)
+        self.data.setdefault[recipient_type.capitalize(), []].append(address)
+
+    def set_subject(self, subject):
+        self.data["Subject"] = subject
+
+    def set_extra_headers(self, headers):
+        self.data.setdefault("Headers", {}).update(headers)
+
+    def set_text_body(self, body):
+        self.data["Text-part"] = body
+
+    def set_html_body(self, body):
+        if "Html-part" in self.data:
+            # second html body could show up through multiple alternatives, or html body + alternative
+            self.unsupported_feature("multiple html parts")
+
+        self.data["Html-part"] = body
+
+    def add_attachment(self, attachment):
+        if attachment.inline:
+            field = "Inline_attachments"
+            name = attachment.cid
+        else:
+            field = "Attachments"
+            name = attachment.name or ""
+        self.data.setdefault(field, []).append({
+            "Content-type": attachment.mimetype,
+            "Filename": name,
+            "content": attachment.b64content
+        })
+
+    def set_tags(self, tags):
+        self.data["Mj-EventPayLoad"] = ','.join(tags)
+
+    def set_track_clicks(self, track_clicks):
+        self.data["Mj-trackclick"] = track_clicks
+
+    def set_track_opens(self, track_opens):
+        self.data["Mj-trackopen"] = track_opens
+
+    def set_template_id(self, template_id):
+        self.data["Mj-TemplateID"] = template_id
+        self.data["MJ-TemplateLanguage"] = True
--- a/anymail/backends/mailjet.py
+++ b/anymail/backends/mailjet.py
@@ -179,8 +179,15 @@ class MailjetPayload(RequestsPayload):
             "content": attachment.b64content
         })
 
+    def set_metadata(self, metadata):
+        self.data["Mj-EventPayLoad"] = metadata
+
     def set_tags(self, tags):
-        self.data["Mj-EventPayLoad"] = ','.join(tags)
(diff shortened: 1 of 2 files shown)


medmunds:
Hey, this is looking great. Thanks for putting it together!

For backend tests, I usually just copy and modify the tests from a similar ESP. It looks like Mailjet's send API uses a straightforward JSON payload, so the Postmark tests could be a good starting point.

anymail/backends/mailjet.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def set_from_email(self, email):
+        self.data["FromEmail"] = email.email
+        if email.name:
+            self.data["FromName"] = email.name
+
+    def add_recipient(self, recipient_type, email):
+        assert recipient_type in ["to", "cc", "bcc"]
+        if email.name:
+            address = "{} <{}>".format(email.name, email.email)
```
medmunds: This *might* have character-escaping issues (if, say, name contains angle brackets or commas).

If the goal is a standard RFC-formatted name/email pair, you can just use `email.address` for both cases. It returns a properly formatted, quoted, and escaped recipient -- e.g., `"Edmunds, Mike" <medmunds@example.com>`. (Haven't checked Mailjet's docs to see what they expect.)

calvin: It expects one of the following formats:
 - "example@example.org"
 - "<example@example.org>"
 - "Doe, John <example@example.org>"

If `email.address` always gives the RFC-compatible name/email pair, I'll just use `email.address`!

Lekensteyn:
FYI: we are evaluating MJ and try to use it with django-anymail. The addition is patched into a "working" state at https://github.com/CodeYellowBV/django-anymail/tree/feature_mailjet

"working" because it can solicit a 200 OK response from MJ, but I could not verify the arrival yet (due to incomplete SPF records).

@calvin Are you still working on this feature?

calvin:
@Lekensteyn 

I'd like to add docs and tests but I'm extremely busy right now.

BTW, I saw your changes and they look very good. Did you receive any mails yet?

Lekensteyn:
@calvin I might also give a shot at adding docs/tests as time permits (but help is of course always welcome :-))

And yes, I managed to receive mail, it was not delivered due to some non-intuitive details of MJ:

 - Subaccounts need a special permission grant from the master account.
 - Even if you can see sub-account keys, you cannot control the Sender addresses/domains.
 - Since it was not obvious that the sub-account key is completely separate from the main one, and the default allowed senders list is empty, mails were not sent. Since the current error reporting is lacking, mails just end up in :blackhole: with no feedback.

In a MJ support ticket, they mentioned that a 3.1 API version is coming which should improve error reporting (public beta for SendAPI 3.1 is estimated for May).

Another issue we noticed is that it is not possible to leave out the FromEmail address (in an attempt to use the address from the template). Doing so will cause an error (400 FromEmail or Sender are required), this will also be addressed in the next API version. Until this new API is ready, you would have to manually pull out the mail from the template.

This seems to be generally useful though, @medmunds is it OK to do another GET request to query the address for a given template which can then be filled into the parameters? When 3.1 is actually released, this hack can then be removed.

Lekensteyn:
Added a quick GET request for the template in CodeYellowBV/django-anymail@2aa2b26d89bc3588d1777486ec96cd7792b07b02. At first I wanted to do a simple `requests.get`, but then figured it would be better to re-use the existing session to save a new TCP connection and TLS handshake. What do you think about this approach? Comments:
- A translation between requests and anymail exceptions is missing
- I considered using "post_to_esp" method with a newly crafted payload class, but that does not seem fitting here as we have no message to send, just a request.
- How can this be tested? Split everything until `json_response = ...` and then mock that?

As for supported Anymail features, do you have a template or list of features where details can be added? It should be pretty complete now (but there are some quirks due to Mailjet limitations).

What do you suggest for testing? Copy an existing file (which?) and adapt it to Mailjet?

medmunds:
@Lekensteyn nice -- clever to reuse the requests session that's already set up. I think you made the right call on skipping post_to_esp.

Not sure what you mean by "translation between requests and anymail exceptions". The main point of using the Anymail exception classes is Anymail uses them to handle Django's weird fail_silently semantics. But you should get them for free since you're calling Anymail's raise_for_status and deserialize_json_response. [... oh, looking through it again, there's one more you'll need to catch -- I'll add a commit comment]

For tests, I'd suggest starting with [test_postmark_backend](https://github.com/anymail/django-anymail/blob/master/tests/test_postmark_backend.py) -- the Postmark send API is pretty similar (in shape) to Mailjet's, from what I can tell. 

Your template sender tests will probably run into RequestsBackendMockAPITestCase's [requests.Session.request patch](https://github.com/anymail/django-anymail/blob/master/tests/mock_requests_backend.py#L30-L42); I *think* you can set mock_request.side_effect to an iterable of return values for the series of request calls. (Unclear how mock [side_effect](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.side_effect) works if [return_value](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.return_value) is also set -- you might need to clear return_value.)

For the list of Anymail message features: any of the backend tests is a good list (either it tests that the feature works, or it tests that it raises AnymailUnsupportedFeature). Or look through the [BasePayload abstract implementation](https://github.com/anymail/django-anymail/blob/master/anymail/backends/base.py#L343-L438) -- anything your payload omits will be unsupported.

And if you have a list of quirks, that's incredibly helpful. (It's usually the hardest part of the ESP-specific docs.) We could handle the docs as a separate PR to avoid holding things up -- just keeping track of the details in this PR is a great start.

I signed up for a Mailjet account, so we can get live integration tests in Travis when you're ready.

Thanks for continuing to push on this.

Lekensteyn:
If I'm not mistaken `self.backend.session.get` could throw a RequestException which needs to be converted into a Anymail exception (AnymailRequestsAPIError?), that is what I meant by translation.

I'll later have a look at adding tests and while doing that fixup the exception handling (test-driven development :)).

medmunds:
Oh, of course -- you're absolutely right.

It would probably be sufficient to just raise an AnymailRequestsAPIError directly in that case, without going through the conversion from requests error like post_to_esp does. (The conversion is there for backwards compatibility with Djrill, where callers might have been looking for a specific requests error. I don't think that would apply to Mailjet users.)

Lekensteyn:
Update: some more bugfixes, tests and fixups of the exception handling on https://github.com/CodeYellowBV/django-anymail/tree/feature_mailjet, here are changes that are not added to this PR yet:
8a64a10 Also handle different method of specifying Sender
fde35e7 mailjet: fix Reply-To support
4eb9227 mailjet: allow setting extra ESP fields
28ae51c mailjet: add metadata and change tags implementation
fc3222a tests: add initial Mailjet backend tests
f45a4b5 tests: add Mailjet test for template responses
0c4d418 Mailjet: convert errors from template API

Tested with: `./runtests.py tests.test_mailjet_backend`

Still needed is an integration test which is harder since there is no testing environment or "sandbox":
https://twitter.com/seanhagen/status/435954163235557377. Furthermore, the API responses isn't that great, it always reports "Sent" even though a mail got held (because the sender address is not whitelisted for example). So for checking the actual status, one would have to do additional API queries (in polling mode).

(Maybe I should open a separate PR such that Travis can automatically check the changes as I push changes?)

medmunds:
@Lekensteyn this is looking amazing! And very complete.

Yes, I think it might be easier to open a new PR on your branch.

Here's how I've been using live integration tests with the other ESPs:

* Most ESPs *don't* offer a sandbox. So the tests run truly live. The recipients are always \@example.com (or SendGrid's special "sink domain" for their tests) to avoid test email arriving in unexpected places.
* I maintain an Anymail test account and validated sender domain for each ESP (that doesn't make this cost-prohibitive). The API keys for those accounts are available in the Travis environment on the master branch. (Travis security blocks those environment vars in most pull requests, so the live tests won't run from your PR until I merge it. But you can set the vars in your environment to your own account for local testing.)
* Most ESPs *don't* return any useful send status from their send APIs. (Email is inherently async, and you have to implement webhooks if you want to know actual send status.) So the live integration tests are mainly making sure the ESP API hasn't changed out from under Anymail (isn't rejecting Anymail payloads). Travis checks this weekly, and it has caught some changes.
* The other big value of the live tests is debugging the initial implementation. While I'm developing, I add myself as a recipient to the [test_all_options](https://github.com/anymail/django-anymail/blob/master/tests/test_sendgrid_integration.py#L61-L89) live test, and set the environment variables to my own ESP account -- and then carefully inspect the mail I receive and the ESP API logs to make sure everything matches the test message. (And running that test usually uncovers some misunderstanding I had about their API.) Same thing for templates and test_merge_data. Just don't forget to remove yourself before committing.

calvin:
Hi guys,

The latest changes from https://github.com/CodeYellowBV/django-anymail/commits/feature_mailjet are merged and pushed to github.

dhrp:
@medmunds, @calvin @Lekensteyn, is there anything left to do? I've successfully implemented this branch in our project and didn't have any problems. Perhaps we can merge it?

I'd be happy to help, if necessary.

Lekensteyn:
@dhrp Functionality-wise it should be ready. One part of the test suite is updated, but an integration test with Mailjet is still needed. It would be great if you could help on that front 😃

medmunds:
@Lekensteyn Travis seems to be failing on some of your functional tests -- looks like mainly around templates. Not sure what might be different from your local environment.

(Sorry, I figured you'd seen those errors after @calvin merged your branch into this PR, but you probably don't get notified. Check the build status at the bottom of the PR on GitHub.)

lorddaedra:
Please refactor to use official https://github.com/mailjet/mailjet-apiv3-python

Lekensteyn:
@medmunds I'll have a look at the tests.

@lorddaedra We investigated that option before, but it is just a very thin wrapper around the REST API which duplicates some parts of anymail. What would be the benefit of using that wrapper?

Lekensteyn:
@medmunds / @calvin Can you fetch the latest branch (with commit a15f1ec933e3c663fbcd15dc4b641740cc48769d) to fix the tests?

medmunds:
@Lekensteyn could I suggest you open a new PR directly from your branch? (I think it'll be easier for you to have direct control. The git history will transfer, so everyone who contributed will still get credit.)

Also, if you haven't recently, it would probably be helpful to rebase your branch onto (this repo's) master before opening the PR.

medmunds:
Superseded by #66.

Issue #65, opened 2017-05-22
Support for MailJet


dhrp:
I see already a couple of other back-ends were requested, but I would love to see [mailjet](https://www.mailjet.com/) integrated. I think mailjet has at least one benefit over most, if not all, others, which is that the templating engine is full-featured and (reasonably) intuitive. On review sites is usually mentioned in the top 7 with transactional e-mail.

I'll look into what it will take to make an implementation myself. -- I also know that [kidig/django-mailjet](https://github.com/kidig/django-mailjet) exists, but seems to miss features. I've been having a hard time settling on a choice (is mailjet reliable enough?) that I would prefer to be able to switch fast.

dhrp:
Just noticed that @calvin started working on a fork for Mailjet! [calvin/django-anymail](/calvin/django-anymail) I'll check out his branch.

medmunds:
@Lekensteyn has moved @calvin's Mailjet implementation forward quite a bit. Check PR #53 for the current state.

lorddaedra:
Check https://github.com/mailjet/mailjet-apiv3-python

Lekensteyn:
@lorddaedra Mailjet's Python API is just a thin interface to their REST API, it does not integrate in anymail (and would thus require rewriting parts if you switch to a different provider at a later moment).

lorddaedra:
@Lekensteyn I mean, it's official library with support, I feel, better to require official libraries (if they exists) and improve them (send PR's) than repeat them in unofficial packages, unix way etc. Just my opinion.

medmunds:
Tasks:
* [X] Implement backend (PR #66)
* [ ] Implement webhooks
* [ ] Docs

medmunds:
Will be released in 0.11

PR #66, opened 2017-06-08
Mailjet backend support (for v0.10)

Lekensteyn:
This PR is based on #53, but rebases on the latest master (0.10) and adds two fixes on top of it (fix for older Mock versions and a fix for test + update for changes in 0.10).

Same caveat holds, the end-to-end tests with Mailjet server is still missing and the documentation is not updated, but I don't have time to address that at the moment.

Closes #53, #65.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -2,6 +2,8 @@ Anymail
 =======
 
 Mike Edmunds
+Calvin Jeong
+Peter Wu
 
 
 Anymail was forked from Djrill, which included contributions from:
--- a/anymail/backends/mailjet.py
+++ b/anymail/backends/mailjet.py
@@ -1,0 +1,124 @@
+from ..exceptions import AnymailRequestsAPIError
+from ..message import AnymailRecipientStatus, ANYMAIL_STATUSES
+from ..utils import get_anymail_setting
+
+from .base_requests import AnymailRequestsBackend, RequestsPayload
+
+
+class EmailBackend(AnymailRequestsBackend):
+    """
+    Mailjet API Email Backend
+    """
+
+    esp_name = "Mailjet"
+
+    def __init__(self, **kwargs):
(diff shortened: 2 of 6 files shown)


medmunds:
Awesome, thanks! I appreciate all your work on this.

It looks like there's still an intermittent failure in [test_api_error_includes_details](https://travis-ci.org/anymail/django-anymail/jobs/240964621). I'm pretty swamped for the next week or so, but can try to take a closer look at that and integration tests next weekend. (And release it, if all goes well.)

medmunds:
I've started putting together a live integration test for this, and ran into some odd behavior where transactional messages apparently don't show up in Mailjet's dashboard, and their MessageIDs don't seem to work with Mailjet's API. Waiting to hear back from MJ support.

medmunds:
OK, as predicted, trying to write the integration tests has uncovered a couple of weirdnesses:

* Mailjet's v3 send API apparently can't handle commas in display-names in the To/Cc/Bcc params. Commas work fine in Recipients...Name params, so this is only a problem when the message is using Cc or Bcc. But the failure is problematic: the message appears to succeed, but isn't actually sent. 
  Not sure whether to try to catch this in Anymail, or just ignore it for now; the beta v3.1 API shouldn't have this issue, as it uses different params for Cc/Bcc. (I'm waiting to hear back from MJ support about the v3 behavior.)
* I'm considering changing Anymail's `tags` from Mailjet's CustomID param to their CustomCampaign param. Mailjet [suggests CustomID](https://dev.mailjet.com/mandrill/#tags) for migrating from Mandrill tags, but I think they misunderstand how tags are typically used. From what I can tell, there's no way in MJ's dashboard to get reporting by CustomID, but you can pull up CustomCampaign-grouped statistics pretty easily.

Lekensteyn:
Looks good, thanks for integrating this!

PR #82, opened 2018-01-22
Mailjet: Upgrade to Send API v3.1

### LINKED ISSUE #81: Upgrade to Mailjet Send API v3.1
Mailjet has quietly released its [v3.1 API](https://dev.mailjet.com/guides/#send-api-v3-1). (Last public notice seems to be the [beta announcement](https://www.mailjet.com/blog/?tag=api).) 

The new API is more logical than the old one, and fixes some problems requiring weird client workarounds—notably, the double API calls currently required to send with a template's From address.

Anymail should upgrade. This will be a breaking change for anyone using `esp_extra` with Mailjet.

medmunds:
[On hold awaiting Mailjet fix for v3.1 Send API regression in template From handling: Mailjet ticket # 830935 2018-01-22.]

Breaking change for `esp_extra` with Mailjet.

Switch from Mailjet's older v3.0 Send API to the latest v3.1 version.

The v3.1 API resolves some limitations/quirks of older API:
* Mailjet now correctly handles commas in recipient display names;
  Anymail no longer works around this using MIME encoded-word syntax.
* Mailjet now supports mixing cc/bcc with merge_data; Anymail no longer
  errors on this case.
* Mailjet now supports using a template's From email and name without
  making an additional API call to retrieve them. (Not currently working
  as documented; awaiting info from Mailjet before releasing.)

But v3.1 adds a new limitation:
* Mailjet supports only a single `reply_to` address.

Closes #81

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -37,6 +37,11 @@ Breaking changes
   (For compatibility with Django 1.11, stay on the Anymail `v7.2 LTS`_
   extended support branch by setting your requirements to `django-anymail~=7.2`.)
 
+* **Mailjet:** Upgrade to Mailjet's newer v3.1 send API. This change should be
+  transparent to most users, but if you are using Anymail's `esp_extra` with Mailjet,
+  you will need to update it for compatibility with the new API. (See
+  `docs <https://anymail.readthedocs.io/en/latest/esps/mailjet/#esp-extra-support>`__.)
+
 * Remove Anymail internal code related to supporting Python 2 and older Django
   versions. This does not change the documented API, but may affect you if your
   code borrowed from Anymail's undocumented internals. (You should be able to switch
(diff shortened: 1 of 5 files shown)


medmunds:
Mailjet API regression still not fixed. Pinging them again for an update.

medmunds:
Closing this until Mailjet fixes their bug with using a template's From address. 

On 2018-01-23, Mailjet support indicated "the cause of the issue was identified and a fix was found in a testing environment". However, as of today, that fix has still not been released.

(We could pull forward the v3.0-API workaround where Anymail makes an extra API call to obtain the template From address, but I was hoping to simplify things by using their v3.1 API as documented. Will re-open if/when they get the fix live.)

medmunds:
2019-02-23: Mailjet v3.1 [docs](https://dev.mailjet.com/guides/#sending-a-basic-email) still read:

> `From`: JSON object ... This property is not mandatory in case you use `TemplateID` and you specified a `From` address for the template. 

... but despite the docs, omitting the `From` field from a template send call generates a "success" response, but the message is not delivered to the recipient (or added to Mailjet activity logs).

medmunds:
Mailjet seems to have (finally) fixed the v3.1 API bug where it wasn't using the template's own From data. We should now be able to update and merge this PR.

medmunds:
Released in Anymail 8.0

Issue #85, opened 2018-01-23
Console or dummy backend for testing


yourcelf:
Feature request: Anymail provides several extensions to stock behavior such as setting `msg.anymail_status` after send,  and firing pre- and post-send signals, none of which happen when using  `django.core.mail.backends.console.EmailBackend`.  As a result, it's difficult to write integration tests for the behavior of signal handlers and status processing code.

A dummy or console backend which at a minimum provides `anymail_status` after send and fires off appropriate pre/post send hooks would make testing code that leverages those features much easier.

medmunds:
Have you tried Anymail's [`test.EmailBackend`](https://anymail.readthedocs.io/en/stable/tips/test_backend/)? It should set the message status and trigger send signals, and has some other features that can be useful for email testing.

yourcelf:
I'd missed that, thanks!  That gets me part of the way there.  For test cases that only create a single message in between database resets, it works fine, but I'm getting blocked by the hard-coded `message_id=1`, which results in unique constraint violations for my send handler which stores message ID's of outgoing messages as an identifier for associating with tracking events.

It would also still be useful to see console output of messages on localhost during development; so I could see a value of a backend that solved both of those issues.

If I find a moment and you're open to the idea, I'll work up a PR.

medmunds:
It would certainly make sense for the test EmailBackend to generate a unique message_id for each recipient. PR welcome.

I'm also open to the idea of a separate console backend. (If you find this is replicating a lot of code from the existing test backend, feel free to refactor to share code if that seems cleaner.) 

BTW, there's an undocumented test EmailBackend feature where you can [set `anymail_test_response`](https://github.com/anymail/django-anymail/blob/v1.2/anymail/backends/test.py#L37-L47) on a message, either to a dict with a `"recipient_status"` key mapping recipient email to `AnymailStatus` ([example](https://github.com/anymail/django-anymail/blob/v1.2/tests/test_send_signals.py#L103-L107)), or to an `AnymailAPIError` instance to simulate a failure. This probably isn't all that useful for your integration tests, but can be helpful for email unit tests.

medmunds:
Released in v1.3. Thanks again!

PR #87, opened 2018-01-24
Un-hardcode message ID in test backend, and add console backend

yourcelf:
Two commits to assist use in development: modify the test backend to use unique message ID's rather than a hardcoded `1`.  Also, add a console backend which provides anymail statuses and signal handling, while printing sent messages to the console.

Re Issue #85.

### DIFF
--- a/anymail/backends/console.py
+++ b/anymail/backends/console.py
@@ -1,3 +1,4 @@
+import uuid
 from django.core.mail.backends.console import EmailBackend as DjangoConsoleBackend
 
 from ..exceptions import AnymailError
@@ -11,6 +12,10 @@ class EmailBackend(AnymailTestBackend, DjangoConsoleBackend):
     """
 
     esp_name = "Console"
+
+    def get_esp_message_id(self, message):
+        # Generate a guaranteed-unique ID for the message
+        return str(uuid.uuid4())
 
     def send_messages(self, email_messages):
         if not email_messages:
--- a/anymail/backends/console.py
+++ b/anymail/backends/console.py
@@ -1,0 +1,38 @@
+from django.core.mail.backends.console import EmailBackend as DjangoConsoleBackend
+
+from ..exceptions import AnymailError
+from .test import EmailBackend as AnymailTestBackend
+
+
+class EmailBackend(AnymailTestBackend, DjangoConsoleBackend):
+    """
+    Anymail backend that prints messages to the console, while retaining
+    anymail statuses and signals.
+    """
+
+    esp_name = "Console"
+
+    def send_messages(self, email_messages):
+        if not email_messages:
+            return
+        msg_count = 0
+        with self._lock:
+            try:
+                stream_created = self.open()
+                for message in email_messages:
+                    try:
+                        sent = self._send(message)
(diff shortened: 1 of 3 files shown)


medmunds:
Thanks for this.

I'm concerned that it's misleading to pull the `message_id` from the email Message-ID header. Most ESPs generate a tracking message_id that's unrelated to the actual header Message-ID, so depending on having that header in the test backend could encourage code that would complicate changing ESPs. (Also, I think a lot of Anymail users don't bother setting their own Message-ID headers, instead letting their ESP generate them during the send.)

Would it work to generate a sequential or random message_id in the test backend? 

Or maybe better for test reproducibility, generate a message_id based on `len(mail.outbox)`, because Django clears that outbox at the same time it resets the test database.

yourcelf:
That makes sense.  The [`EmailMultiAlternatives.message()`](https://github.com/django/django/blob/master/django/core/mail/message.py#L251-L276) generated message ID gets regenerated on every call to `.message()`; but `mail.outbox` stores the `EmailMultiAlternatives` instance itself and not the serialized output of `.message()`.  As a result, a test can't use the generated output of `.message()` to predict the message ID that you'd get on a future call to `.message()`.  That's why I provided the `Message-ID` header in the test case, to bypass that auto-generation.

Moving away from `.message()`'s implementation and doing something more deterministic could help avoid confusion in the test case, sure.  Basing it on `len(mail.outbox)` would work in unit test contexts, but wouldn't ideal for localhost development, at least as long as you expect mesage IDs to be unique.  Everytime you restart the dev server, `len(mail.outbox)` would be reset to 0, and then you'd be reusing message IDs again (and potentially running afoul of unique constraints).

Perhaps the implementation for generating a message ID really needs to be different between test and console/localhost development cases, where test message ID's are deterministic based on something like `len(mail.outbox)`, but console message IDs are based on clock time like the current implementation.  Would be easy to implement a `get_message_id` method on the test backend, and have the console backend which derives from it override that.

medmunds:
Yeah, none of the Anymail backends call `.message()`—the RFC-x822 format it generates is really only useful for communicating with SMTP servers, not ESP's REST APIs. (Also, that string won't include any of Anymail's custom props like tags or metadata. You might find it more helpful to print the Test backend's payload.params dict, which is everything Anymail knows about the message.)

If you want to do something different in the console backend to facilitate local development, you could override `AnymailBaseBackend.pre_send_message()` and/or `.post_send_message()`. (Be sure to chain to super, or the signal handlers won't get run.)

yourcelf:
I've added a commit to use different strategies for acquiring a message ID on the `test` and `console` backends.

The `test` backend uses the array position of the message in `mail.outbox`.  This makes the message ID deterministic and predictable across tests without a need to specify the ID in advance, but ensures that ID's for multiple messages will remain unique within a single test.

The `console` backend just uses a uuid, ensuring global uniqueness for localhost development.  This ensures that it will never collide, and keeps the ESP message ID semantically distinct from the SMTP message ID.

medmunds:
Thanks!

medmunds:
Oops, closes #85.

PR #89, opened 2018-02-01
[WIP] Responsys backend

atif:
[#154524653] Add a back-end provider to support responsys emails

### DIFF
--- a/anymail/backends/responsys.py
+++ b/anymail/backends/responsys.py
@@ -1,0 +1,90 @@
+import requests
+
+from ..utils import get_anymail_setting
+from ..message import AnymailRecipientStatus
+from .base_requests import AnymailRequestsBackend, RequestsPayload
+
+class EmailBackend(AnymailRequestsBackend):
+    """
+    Responsys API Email Backend
+    """
+
+    esp_name = "Responsys"
+
+    def __init__(self, **kwargs):
+        """Init options from Django settings"""
+        esp_name = self.esp_name
+
+        username = get_anymail_setting('username', esp_name=esp_name, kwargs=kwargs, default=None, allow_bare=True)
+        password = get_anymail_setting('password', esp_name=esp_name, kwargs=kwargs, default=None, allow_bare=True)
+
+        api_url = get_anymail_setting('api_url', esp_name=esp_name, kwargs=kwargs,
+                                        default='/rest/api/v1.3/campaigns/test_email_alerts/email')
+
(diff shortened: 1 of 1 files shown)


medmunds:
Thanks for the contribution. I'm assuming this is a work-in-progress, based on all the stubbed-out payload methods. I'll wait for your updates, but just speak up if there are any areas where you'd like some help moving it forward.

Couple of thoughts:
* I'm guessing you used `pass` in the payload methods as a placeholder until you get around to implementing those properties. But if there end up being some Anymail/Django message properties that Responsys truly doesn't support, you'd want to use `self.unsupported_feature("...property name or description...")` to warn users they don't work.
* It looks like you've hard-coded [your own campaignName](https://github.com/anymail/django-anymail/pull/89/files#diff-a1cd28ed9f0135ce500f7241347b599fR22) in the api url. The Mailgun backend has an example of [using message properties to alter the url](https://github.com/anymail/django-anymail/blob/v1.2/anymail/backends/mailgun.py#L73-L79) which might be relevant. (From what I can tell, Anymail's [`template_id`](https://anymail.readthedocs.io/en/stable/sending/templates/#anymail.message.AnymailMessage.template_id) is probably the equivalent to [Responsys's campaignName](https://docs.oracle.com/cloud/latest/marketingcs_gs/OMCED/op-rest-api-v1.3-campaigns-%7BcampaignName%7D-email-post.html#request), but I'm not at all familiar with Responsys.)

atif:
@medmunds Thanks for your valuable feedback. Yes, it's a work in progress. Will definitely ping you in case I need more guidance! :)

medmunds:
@atif sounds good.

BTW, I finally added some [developer notes on adding ESPs](https://github.com/anymail/django-anymail/blob/master/ADDING_ESPS.md) yesterday, which might be helpful. (Though looks like you're figuring out Anymail's internals pretty well on your own.)

atif:
@medmunds Thanks for adding the notes!

medmunds:
@atif are you still working on this? Let me know if there's anything I can do to help. 

(Unfortunately, I don't have access to Responsys—it looks like it's an enterprise-sales-only type product. If there's a way I can get a test account, let me know.)

atif:
@medmunds No, we shifted to Amazon SES for now. Yes, it's an enterprise solution and doesn't really map to django-anymail as it doesn't fall truly in the transactional ESPs.

PR #93, opened 2018-02-17
[WIP] Feature integration sendinblue

### LINKED ISSUE #84: SendinBlue
Please add support for [SendinBlue](https://www.sendinblue.com/).

Thanks!

RignonNoel:
| Q                                                      | R
| ------------------------------------------ | -------------------------------------------
| Type of contribution ?                      | Feature
| Tickets (_issues_) concerned               | fixed #84

---

I've begin an integration of `SendinBlue` with the new API v3, i took the choice to directly call the API and to not use the official python client implementation as recommended by `django-anymail` documentation.

Please feel free to review my code, contribute or just help me with some tips or good idea.

I will continue this work soon 👍 
 
##### What have you changed ?
-  [x] Link the issue to this PR
-  [x] Add my name in `AUTHORS.txt`
-  [x] Create the Backend and Payload classes
-  [x] Add or modify the attached tests
-  [x] Add or modify the attached documentation
-  [x] Test with an account of the ESP (Thank for the great integration testing ❤️ )
-  [x] Add templated transactional email support
-  [x] Add tag support

##### How did you change it ?
I've based my work on [this awaiting PR for Mailjet](https://github.com/anymail/django-anymail/blob/da25eb03b64d28d0a4effcf4fe014cd86ac9a7fd/anymail/backends/mailjet.py) and the `Sendgrid` integration.

##### References

 1. [SendinBlue's general documentation for API v3](https://developers.sendinblue.com/docs/getting-started)
 2. [SensinBlue's dynamic test for API v3](https://developers.sendinblue.com/v3.0/reference#sendtransacemail-1)
 3. [SendinBlue's official client API v3](https://github.com/sendinblue/APIv3-python-library)

### DIFF
--- a/anymail/backends/sendinblue.py
+++ b/anymail/backends/sendinblue.py
@@ -223,3 +223,7 @@ class SendinBluePayload(RequestsPayload):
 
     def set_merge_global_data(self, merge_global_data):
         self.data['attributes'] = merge_global_data
+
+    def set_metadata(self, metadata):
+        # SendinBlue expects a single string payload
+        self.data['headers']["X-Mailin-custom"] = self.serialize_json(metadata)
--- a/anymail/backends/sendinblue.py
+++ b/anymail/backends/sendinblue.py
@@ -191,9 +191,6 @@ class SendinBluePayload(RequestsPayload):
     def set_template_id(self, template_id):
         self.template_id = template_id
 
-    def set_template_id(self, template_id):
-        self.template_id = template_id
-
     def set_text_body(self, body):
         if body:
             self.data['textContent'] = body
--- a/anymail/backends/sendinblue.py
+++ b/anymail/backends/sendinblue.py
@@ -1,0 +1,224 @@
+from requests.structures import CaseInsensitiveDict
+
+from .base_requests import AnymailRequestsBackend, RequestsPayload
+from ..exceptions import AnymailRequestsAPIError
+from ..message import AnymailRecipientStatus
+from ..utils import get_anymail_setting, parse_address_list
+
+
+class EmailBackend(AnymailRequestsBackend):
+    """
+    SendinBlue v3 API Email Backend
+    """
+
+    esp_name = "SendinBlue"
+
+    def __init__(self, **kwargs):
+        """Init options from Django settings"""
+        esp_name = self.esp_name
+        self.api_key = get_anymail_setting(
+            'api_key',
+            esp_name=esp_name,
+            kwargs=kwargs,
+            allow_bare=True,
+        )
+        api_url = get_anymail_setting(
+            'api_url',
+            esp_name=esp_name,
+            kwargs=kwargs,
+            default="https://api.sendinblue.com/v3",
+        )
+        if not api_url.endswith("/"):
+            api_url += "/"
+        super(EmailBackend, self).__init__(api_url, **kwargs)
+
+    def build_message_payload(self, message, defaults):
+        return SendinBluePayload(message, defaults, self)
+
+    def raise_for_status(self, response, payload, message):
+        if response.status_code < 200 or response.status_code >= 300:
+            raise AnymailRequestsAPIError(
+                email_message=message,
+                payload=payload,
+                response=response,
+                backend=self,
+            )
+
+    def parse_recipient_status(self, response, payload, message):
+        # SendinBlue doesn't give any detail on a success
+        # https://developers.sendinblue.com/docs/responses
+        message_id = None
+
+        if response.content != b'':
+            parsed_response = self.deserialize_json_response(response, payload, message)
+            try:
+                message_id = parsed_response['messageId']
+            except (KeyError, TypeError):
+                raise AnymailRequestsAPIError("Invalid SendinBlue API response format",
+                                              email_message=message, payload=payload, response=response,
+                                              backend=self)
+
+        status = AnymailRecipientStatus(message_id=message_id, status="queued")
+        return {recipient.addr_spec: status for recipient in payload.all_recipients}
+
+
+class SendinBluePayload(RequestsPayload):
+
+    def __init__(self, message, defaults, backend, *args, **kwargs):
+        self.all_recipients = []  # used for backend.parse_recipient_status
+        self.template_id = None
+
+        http_headers = kwargs.pop('headers', {})
+        http_headers['api-key'] = backend.api_key
+        http_headers['Content-Type'] = 'application/json'
+
+        super(SendinBluePayload, self).__init__(message, defaults, backend, headers=http_headers, *args, **kwargs)
+
+    def get_api_endpoint(self):
+        if self.template_id:
+            return "smtp/templates/%s/send" % (self.template_id)
+        else:
+            return "smtp/email"
+
+    def init_payload(self):
+        self.data = {  # becomes json
+            'headers': CaseInsensitiveDict()
+        }
+
+    def serialize_data(self):
+        """Performs any necessary serialization on self.data, and returns the result."""
+
+        headers = self.data["headers"]
+        if "Reply-To" in headers:
+            # Reply-To must be in its own param
+            reply_to = headers.pop('Reply-To')
+            self.set_reply_to(parse_address_list([reply_to]))
+        if len(headers) > 0:
+            self.data["headers"] = dict(headers)  # flatten to normal dict for json serialization
+        else:
+            del self.data["headers"]  # don't send empty headers
+
+        # SendinBlue use different argument's name if we use template functionality
+        if self.template_id:
+            data = self._transform_data_for_templated_email(self.data)
+        else:
+            data = self.data
+
+        return self.serialize_json(data)
+
+    def _transform_data_for_templated_email(self, data):
+        """
+        Transform the default Payload's data (used for basic transactional email) to
+        the data used by SendinBlue in case of a templated transactional email.
+        :param data: The data we want to transform
+        :return: The transformed data
+        """
+        if 'subject' in data:
+            self.unsupported_feature("overriding template subject")
+        if 'subject' in data:
+            self.unsupported_feature("overriding template from_email")
+        if 'textContent' in data or 'htmlContent' in data:
+            self.unsupported_feature("overriding template body content")
+
+        transformation = {
+            'to': 'emailTo',
+            'cc': 'emailCc',
+            'bcc': 'emailBcc',
+        }
+        for key in data:
+            if key in transformation:
+                new_key = transformation[key]
+                list_email = list()
+                for email in data.pop(key):
+                    if 'name' in email:
+                        self.unsupported_feature("display names in (%r) when sending with a template" % key)
+
+                    list_email.append(email.get('email'))
+
+                data[new_key] = list_email
+
+        if 'replyTo' in data:
+            if 'name' in data['replyTo']:
+                self.unsupported_feature("display names in (replyTo) when sending with a template")
+            data['replyTo'] = data['replyTo']['email']
+
+        return data
+
+    #
+    # Payload construction
+    #
+
+    @staticmethod
+    def email_object(email):
+        """Converts EmailAddress to SendinBlue API array"""
+        email_object = dict()
+        email_object['email'] = email.addr_spec
+        if email.display_name:
+            email_object['name'] = email.display_name
+        return email_object
+
+    def set_from_email(self, email):
+        self.data["sender"] = self.email_object(email)
+
+    def set_recipients(self, recipient_type, emails):
+        assert recipient_type in ["to", "cc", "bcc"]
+        if emails:
+            self.data[recipient_type] = [self.email_object(email) for email in emails]
+            self.all_recipients += emails  # used for backend.parse_recipient_status
+
+    def set_subject(self, subject):
+        if subject != "":  # see note in set_text_body about template rendering
+            self.data["subject"] = subject
+
+    def set_reply_to(self, emails):
+        # SendinBlue only supports a single address in the reply_to API param.
+        if len(emails) > 1:
+            self.unsupported_feature("multiple reply_to addresses")
+        if len(emails) > 0:
+            self.data['replyTo'] = self.email_object(emails[0])
+
+    def set_extra_headers(self, headers):
+        for key in headers.keys():
+            self.data['headers'][key] = headers[key]
+
+    def set_tags(self, tags):
+        if len(tags) > 0:
+            self.data['headers']["X-Mailin-tag"] = tags[0]
+            if len(tags) > 1:
+                self.unsupported_feature('multiple tags (%r)' % tags)
+
+    def set_template_id(self, template_id):
+        self.template_id = template_id
+
+    def set_text_body(self, body):
+        if body:
+            self.data['textContent'] = body
+
+    def set_html_body(self, body):
+        if body:
+            if "htmlContent" in self.data:
+                self.unsupported_feature("multiple html parts")
+
+            self.data['htmlContent'] = body
+
+    def add_attachment(self, attachment):
+        """Converts attachments to SendinBlue API {name, base64} array"""
+        att = {
+            'name': attachment.name or '',
+            'content': attachment.b64content,
+        }
+
+        if attachment.inline:
+            self.unsupported_feature("inline attachments")
+
+        self.data.setdefault("attachment", []).append(att)
+
+    def set_esp_extra(self, extra):
+        self.data.update(extra)
+
+    def set_merge_data(self, merge_data):
+        """SendinBlue doesn't support special attributes for each recipient"""
+        self.unsupported_feature("merge_data")
+
+    def set_merge_global_data(self, merge_global_data):
+        self.data['attributes'] = merge_global_data
--- a/anymail/backends/sendinblue.py
+++ b/anymail/backends/sendinblue.py
@@ -140,6 +140,7 @@ class SendinBluePayload(RequestsPayload):
         if 'replyTo' in data:
             if 'name' in data['replyTo']:
                 self.unsupported_feature("display names in (replyTo) when sending with a template")
+
             data['replyTo'] = data['replyTo']['email']
 
         return data
@@ -186,6 +187,9 @@ class SendinBluePayload(RequestsPayload):
             self.data['headers']["X-Mailin-tag"] = tags[0]
             if len(tags) > 1:
                 self.unsupported_feature('multiple tags (%r)' % tags)
+
+    def set_template_id(self, template_id):
+        self.template_id = template_id
 
     def set_template_id(self, template_id):
         self.template_id = template_id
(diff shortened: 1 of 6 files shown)


medmunds:
Very nice start—thanks for this!

I'll add a few specific notes inline, but overall this is looking good.

Some of their API documentation seems a little, um, under-specified, so there will likely be some experimentation required to get all the details right. (In some cases, their [v2 docs](https://apidocs.sendinblue.com/tutorial-sending-transactional-email/) might provide some direction; although the v3 API is different, I'm guessing it shares a lot of the same underlying platform capabilities and limitations.)

After you get the basic transactional send working, there are some additional Anymail send features that should probably be supported:
* `template_id` and `merge_data`: SiB's [transactional-template](https://developers.sendinblue.com/v3.0/reference#sendtemplate) sending
* `tags` and `metadata`: looks like maybe special X-Mailin-custom and X-Mailin-tag headers can be used for these (based on v2 docs)

anymail/backends/sendinblue.py
```diff
@@ -0,0 +1,144 @@
...
+                payload=payload,
+                response=response,
+                backend=self,
+            )
+
+    def parse_recipient_status(self, response, payload, message):
+        # SendinBlue doesn't give any detail on a success
+        # https://developers.sendinblue.com/docs/responses
+
+        return dict()
```
medmunds: This *must* return a dict mapping recipient email address (addr_spec) to an AnymailRecipientStatus for each message recipient. Since the best we know here is "probably sent", you could use [code like this](https://github.com/anymail/django-anymail/blob/v1.4/anymail/backends/mailgun.py#L54-L56) to mark all recipients "queued".

The AnymailRecipientStatus.message_id is some sort of tracking ID that can be used to correlate this send with later webhook status events. It's not documented, but a quick test suggests SiB's response is actually json, and *does* include a usable tracking id: `{"messageId":"<201802182116.25729519001@smtp-relay.mailin.fr>"}`.

anymail/backends/sendinblue.py
```diff
@@ -0,0 +1,144 @@
...
+        return email_object
+
+    def set_from_email(self, email):
+        self.data["sender"] = self.email_object(email)
+
+    def set_recipients(self, recipient_type, emails):
+        assert recipient_type in ["to", "cc", "bcc"]
+        if emails:
+            # Normally, exactly one "personalizations" entry for all recipients
+            # (Exception: with merge_data; will be burst apart later.)
```
medmunds: (This comment doesn't apply to SiB, I think.)

anymail/backends/sendinblue.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def set_reply_to(self, emails):
+        # SendinBlue only supports a single address in the reply_to API param.
+        if len(emails) > 1:
+            self.unsupported_feature("multiple reply_to addresses")
+        if len(emails) > 0:
+            self.data['replyTo'] = self.email_object(emails[0])
+
+    def set_extra_headers(self, headers):
+        self.data['headers'] = [header for header in headers]
```
medmunds: SiB expects an "object" here, but doesn't say much more. I'm guessing it's a mapping of header field-name to value, which is what the `headers` param already is. So you can probably just do `self.data['headers'] = headers`.

anymail/backends/sendinblue.py
```diff
@@ -0,0 +1,144 @@
...
+    def set_html_body(self, body):
+        if body is not None:
+            if "htmlContent" in self.data:
+                self.unsupported_feature("multiple html parts")
+
+            self.data['htmlContent'] = body
+
+    def add_attachment(self, attachment):
+        """Converts attachments to SendinBlue API {name, base64} array"""
+
```
medmunds: I can't find any info on SiB support for inline attachments/inline images. (May need to contact them for clarification.) If they truly don't support inlines, we'd need `if attachment.inline:   self.unsupported_feature("inline attachments")` here.

(Also, I'm really surprised they don't have a way to specify the attachment content-type; I'm wondering if they always just guess that from the name. If so, that becomes a "quirks and limitations" entry for the docs.)

anymail/backends/sendinblue.py
```diff
@@ -0,0 +1,144 @@
...
+        att = {
+            'name': attachment.name,
+            'content': attachment.b64content,
+        }
+
+        self.data.setdefault("attachment", []).append(att)
```
medmunds: Every payload should implement `set_esp_extra` to allow overrides and esp-specific parameters. Usually just dict.update into self.data works fine. (Anymail utils.update_deep can also be used.) [Example](https://github.com/anymail/django-anymail/blob/v1.4/anymail/backends/mailgun.py#L198-L199).

anymail/backends/sendinblue.py
RignonNoel: Exactly responses i've on my tests look like this one :
```
{
    'code': 'success',
    'message': 'Email sent successfully',
    'data': {
        'message-id': '<201802182329.73817096184@smtp-relay.sendinblue.com>'
    }
}
```

Thanks for the link on the mailgun example. It was one of my point on my todo-list of research 👍

anymail/backends/sendinblue.py
RignonNoel: On the API v2 we have this little part of the documentation.
Maybe the v3 keep it functionnal.. We will need to test i guess.

|Input Parameter | Description|
| -- | -- |
| html | Body of the message (HTML version) [Mandatory]. To send inline images, use < img src="{YourFileName.Extension}" alt="image" border="0" >, the src attribute value inside {} (curly braces) should be same as the filename used in inline_image parameter |
| inline_image | Pass your inline image/s filename & its  base64 encoded chunk data as an associative array. Example:  array(“YourFileName.Extension”=>”Base64EncodedChunkData”). You can  use commas to separate multiple inline images [Optional]. |

anymail/backends/sendinblue.py
RignonNoel: The documentation of v2 have more detail on the `headers` parameter :

|Input Parameter | Description|
| -- | -- |
| headers | The headers will be sent along with the mail  headers in original email. Example: array(“Content-Type”=>”text/html;  charset=iso-8859-1″). You can use commas to separate multiple headers  [Optional]. |

I guess it's just the same thing.

anymail/backends/sendinblue.py
RignonNoel: Done. I've found no one information for the extra header on the v3 so i've not included some check in the code.

> Note:
“X-Mailin-IP” will not be sent along with the mail headers in original email.
“X-param1″, “X-param2″, “X-Mailin-custom” and “X-Mailin-Tag” will be sent along with the mail headers in original email.
“X-Mailin-custom” and “X-Mailin-Tag” will also be sent along with the JSON response of webhooks, when some event occurs. Also, the “X-Mailin-Tag” will be received as only “tag” in the JSON response.

[API v2 reference](https://apidocs.sendinblue.com/tutorial-sending-transactional-email/)

RignonNoel:
Hi @medmunds and thanks a lot for all your comments! 

---

I've check the python client API for the v3 and i find [the part of the code](https://github.com/sendinblue/APIv3-python-library/blob/2c74a142582f23f265a2e3f433e5516a4a13a3cb/sib_api_v3_sdk/api/smtp_api.py#L1086-L1182) about transactional email without template. Unfortunately, no trace of the `inline_image` feature of the v2... 

I will ask the tech team of SendinBlue by email tomorrow.

---

For the `template` feature i'm not sure what's the best architecture to do it considering the fact that SendinBlue doesn't use the same url for templated email.

```
https://api.sendinblue.com/v3/smtp/templates/templateId/send
```

Do you have some documentation about Django's email backend i can read to understand how `set_template_id` is called ? I do not understand all the call stack used to make anymail's backend works. 

Otherwise it does not matter i'm just going to read django's documentation directly 😃

medmunds:
> For the template feature i'm not sure what's the best architecture to do it considering the fact that SendinBlue doesn't use the same url for templated email.

You can add logic to your Payload.get_api_endpoint. Examples in the [Postmark backend](https://github.com/anymail/django-anymail/blob/v1.4/anymail/backends/postmark.py#L109-L114) and  [Mandrill backend](https://github.com/anymail/django-anymail/blob/v1.4/anymail/backends/mandrill.py#L74-L78).

RignonNoel:
I just worked on the integration of templated email but their API seems really inconsistent on some part :

 - Fields `to`, `cc` and `bcc` are renamed `emailTo`, `emailCc` and `emailBcc` in case you use a `template_id`
 - These fields doesn't take the same kind of object inside:
   - `to` : `[{"email": "to1@example.com", "name": "My name"}]`
   - `emailTo`: `["to1@example.com"]`
 - `replyTo` keep the same name but do not have the same content also

Considering this really poor conception of their API i've made a custom function to rename fields and change the content. My custom function is called in the `serialize_data()` in case we have a `template_id`.

---

I've thought about manage two variable `self.data` and `self.data_template` and keep them both updated on EACH function but i was not sure about which part of the system have access to `self.data` since the variable doens't use the `private` practice of python (ie: `self._data`).

The actual implementation i've made required to change the new custom function in case we change the `self.data` structure (ex: new SiB's API documentation). It's not a big problem since v3 is officially release, but it's not the best way to do it.

medmunds:
@RignonNoel awesome! Sorry for the delay responding.

The only significant piece missing for the backend right now is backend (non-integration) tests. These mock out requests to verify that Anymail is trying to send the expected parameters to SendinBlue's API, for a variety of different cases. I usually just copy the tests for an ESP with a similar API (e.g., [test_mailjet_backend.py](https://github.com/anymail/django-anymail/blob/feature_mailjet_v3.1/tests/test_mailjet_backend.py)) and modify them. Let me know if that's enough to go on, or if I can provide any help.

Re: their template API using a totally different format for to/cc/etc., that sort of inconsistency is more common than you'd think/hope. And `serialize_data()` is the right place to fix it up. (Search for "late binding" or "late-bound" in the other backends to see [several examples](https://github.com/anymail/django-anymail/search?q=late&type=Code&utf8=%E2%9C%93) of this.)

You can certainly transform the existing data in serialize_data. Another approach I often use is to just [stash the raw values](https://github.com/anymail/django-anymail/blob/v1.4/anymail/backends/mailjet.py#L182-L186) in the payload (outside the data prop), and then [process them into data](https://github.com/anymail/django-anymail/blob/v1.4/anymail/backends/mailjet.py#L91-L107) in serialize_data, once I know what format is needed.

Once the backend tests are in place, I think we can merge this. The webhooks can come later, as a separate PR. (Also, I signed up for a SendinBlue account, so we'll be able to run your live integration tests in the Travis builds.)

anymail/backends/sendinblue.py
```diff
@@ -0,0 +1,200 @@
...
+        """
+        transformation = {
+            'to': 'emailTo',
+            'cc': 'emailCc',
+            'bcc': 'emailBcc,'
+        }
+        for key in self.data.keys():
+            if key in transformation.keys():
+                new_key = transformation[key]
+                self.data[new_key] = [email.get('email') for email in self.data.pop(key)]
```
medmunds: I think this is going to lose the recipient's display_name when sending a template email.

What you probably want here is the fully-formatted [`email.address`](https://github.com/anymail/django-anymail/blob/v1.4/anymail/backends/mailjet.py#L91-L107) from the original EmailAddress objects passed to `set_recipients`.

RignonNoel: SendinBlue doesn't support `display_name` on templated email :unamused: ...

https://developers.sendinblue.com/v3.0/reference#sendtemplate
```
curl --request POST \
  --url https://api.sendinblue.com/v3/smtp/templates/templateId/send \
  --data '{"emailTo":["to1@example.com","to2@example.com"]}'
```

RignonNoel:
Hi @medmunds, No problem for the delay we all have a life on the side! 

I just added some backend test for sendinblue. I hope it do the work for you, i tried to highlight all unsupported feature and limitation (no metadata, only one tag, ..)

I just have a very empty test for `template_id`. Have you a way to test that on this context ? For the moment it just test to send and check if no error are trigger but it's not the best check i've seen in my life :/

tests/test_sendinblue_backend.py
```diff
@@ -0,0 +1,444 @@
...
+        self.message.track_opens = False
+        with self.assertRaises(AnymailUnsupportedFeature):
+            self.message.send()
+
+    def test_template_id(self):
+        # SendinBlue use incremental ID to identify templates
+        self.message.template_id = "12"
+        self.message.send()
+
+        # We can't verify if the Url has changed
```
medmunds: `self.assert_esp_called('/v3/smtp/templates/12/send')` will check the URL. (See RequestsBackendMockAPITestCase.) 

This would also be a good place to verify that all the API parameters are transformed as expected for the template API call. ("to" --> "emailTo", etc.)

tests/test_sendinblue_backend.py
```diff
@@ -0,0 +1,444 @@
...
+        message = mail.EmailMessage(from_email='from@example.com', to=['to@example.com'])
+        message.template_id = "9"
+        message.send()
+        data = self.get_api_call_json()
+        self.assertNotIn('htmlcontent', data)
+        self.assertEquals(data['textContent'], '')  # neither text nor html body
+        self.assertNotIn('subject', data)
+
+    def test_merge_data(self):
+        self.message.from_email = 'from@example.com'
```
medmunds: For most backends, this is a combined test for `message.merge_data` and `message.merge_global_data`. Since SendinBlue only supports the global variant, it'd be good to split that out (or roll it into the template test earlier), and verify that merge_global_data ends up as SiB "attributes". [Example from Postmark tests](https://github.com/anymail/django-anymail/blob/v1.4/tests/test_postmark_backend.py#L362-L377)

tests/test_sendinblue_backend.py
```diff
@@ -0,0 +1,444 @@
...
+        Options not specified by the caller should be omitted entirely from
+        the API call (*not* sent as False or empty). This ensures
+        that your ESP account settings apply by default.
+        """
+        self.message.send()
+        data = self.get_api_call_json()
+        self.assertNotIn('attachment', data)
+        self.assertNotIn('tag', data)
+        self.assertNotIn('headers', data)
+        self.assertNotIn('send_at', data)
```
medmunds: This should assertNotIn for all the optional ESP API params. So, it's not necessary to check `'send_at'` (not a SiB param), but it'd be good to check `'replyTo'` and maybe `'attributes'`.

tests/test_sendinblue_backend.py
```diff
@@ -0,0 +1,444 @@
...
+    # noinspection PyUnresolvedReferences
+    def test_send_attaches_anymail_status(self):
+        """ The anymail_status should be attached to the message when it is sent """
+        # the DEFAULT_RAW_RESPONSE above is the *only* success response SendinBlue returns,
+        # so no need to override it here
+        msg = mail.EmailMessage('Subject', 'Message', 'from@example.com', ['to1@example.com'],)
+        sent = msg.send()
+        self.assertEqual(sent, 1)
+        self.assertEqual(msg.anymail_status.status, {'queued'})
+        self.assertEqual(msg.anymail_status.recipients['to1@example.com'].status, 'queued')
```
medmunds: This should also check that the `msg.anymail_status.message_id` and `msg.anymail_status.recipients['to1@example.com'].message_id` get set to the message_id from SiB's API response.

medmunds:
@RignonNoel backend tests look great—thanks! I'll add inline notes on testing template_id and a few other things.

I experimented a little more with the SendinBlue transactional APIs, and found a few things:
* You can set an X-Mailin-custom header to an arbitrary string, and it will be passed to webhooks. This would be a good way to implement Anymail's [metadata](https://anymail.readthedocs.io/en/stable/sending/anymail_additions/#anymail.message.AnymailMessage.metadata) extension, by passing it as JSON ([example](https://github.com/anymail/django-anymail/blob/v1.4/anymail/backends/mailjet.py#L224-L226)). The exact capitalization of the header name seems to matter, and SiB's docs aren't always consistent. Lowercase "custom" is the one that worked for me. (Ditto X-Mailin-tag, btw. X-Mailin-Tag with a capital T doesn't get passed to webhooks.)
* Attachment content-type is determined from the filename extension. (And trying to send an attachment without a name or a name without an extension generates a 400 response with an "Unsupported file format" message. I don't think the code needs to do anything special for this, but it's worth noting in the docs.)
* SendinBlue has a bunch of limitations that might not be expected coming from other ESPs (e.g., can't send inline images, can't specify display-names when sending a template, etc.). I'll add inline notes for the other ones I found. In all these cases, the payload needs to call `self.unsupported_feature("...")`. (Issuing errors on unsupported features is important for users trying to switch ESPs, in case they're counting on something that won't work.)

Let me know if you want to take a crack at those, otherwise I can fill them in.

Also, you're the first person other than me to fully implement a new backend for Anymail, so a huge thanks! I've learned a lot from this about places I should improve the internal documentation and porting notes, and I really appreciate the effort you put into powering through what's there now.

anymail/backends/sendinblue.py
```diff
@@ -0,0 +1,205 @@
...
+                backend=self,
+            )
+
+    def parse_recipient_status(self, response, payload, message):
+        # SendinBlue doesn't give any detail on a success
+        # https://developers.sendinblue.com/docs/responses
+        message_id = None
+
+        if response.content != b'':
+            parsed_response = self.deserialize_json_response(response, payload, message)
```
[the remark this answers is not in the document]
medmunds: Uh, sorry—ignore this comment. deserialize_json_response worries about that for you.

tests/test_sendinblue_backend.py
RignonNoel: Done

tests/test_sendinblue_backend.py
```diff
@@ -0,0 +1,444 @@
...
+    def test_unicode_attachment_correctly_decoded(self):
+        self.message.attach(u"Une pièce jointe.html", u'<p>\u2019</p>', mimetype='text/html')
+        self.message.send()
+        attachment = self.get_api_call_json()['attachment'][0]
+        self.assertEqual(attachment['name'], u'Une pièce jointe.html')
+        self.assertEqual(b64decode(attachment['content']).decode('utf-8'), u'<p>\u2019</p>')
+
+    def test_embedded_images(self):
+        # SendinBlue doesn't support inline image
+        # inline image are just added as a content attachment
```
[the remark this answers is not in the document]
RignonNoel: Done

tests/test_sendinblue_backend.py
RignonNoel: Done

tests/test_sendinblue_backend.py
RignonNoel: Done

tests/test_sendinblue_backend.py
RignonNoel: Done

RignonNoel:
@medmunds Thanks a lot for all your comments! It really help me to understand some part of anymail (i really look forward to use it now!! :smile: ).

 - I rebased all my commits to fix your inline comments
 - I added some details on SendinBlue documentation (I did it concisely because I still write pretty bad English...)
 - I added metadata support in the Payload and refactor attached test

Let me know if you have some other comments!

medmunds:
@RignonNoel outstanding, thanks! Merging now. I'll aim to release it in the next few days.

Let me know if you want to take a crack at SendinBlue webhooks.

medmunds:
Just a followup: SendinBlue support has confirmed that their "V3 [API does] not support inline images."

Issue #54, opened 2017-03-05
Amazon SES


mikenunez:
Please, add an easy support for amazon SES or https://pepocampaigns.com/

Thank you

medmunds:
[Anyone else wanting Amazon SES support in Anymail, please use GitHub's "thumbs up" [reaction](https://github.com/blog/2119-add-reactions-to-pull-requests-issues-and-comments) to vote for this issue.]

By the way, there are already at least two packages that implement SES EmailBackends for Django: [django-ses](https://pypi.python.org/pypi/django-ses) and [django-amazon-ses](https://pypi.python.org/pypi/django-amazon-ses). Is there some reason neither of those meets your needs?

It would help me to understand if there are some specific Anymail features you're looking for from an SES integration. (The last time I checked, SES didn't offer most of the "value-added transactional" features Anymail exposes from its current supported ESPs.)

thongly:
@medmunds - Is anyone working on this branch, and is there any way we can help?

medmunds:
@thongly coincidentally, I was just looking at SES again yesterday. Looks like Amazon recently added a bunch of tracking and notification features that make SES more similar to the other ESPs in Anymail.

One thing that would still be helpful (from any of the upvoters reading this) is understanding why django-ses or django-amazon-ses  isn't a good fit -- I don't want to duplicate those projects unnecessarily.

It'll be at least a couple of weeks until I get a chance to look at SES again. If anyone wants to take a crack at it before then, the help is always welcome! Just add a note to this issue when you start, and feel free to post work-in-progress/incomplete PRs for collaboration.

Dump of what I (think I) learned about SES yesterday...

* Simpler SES SendEmail API doesn't support attachments, so probably use SendRawEmail instead -- takes a fully composed MIME message. (Good news: Django's EmailMessage objects are really good at serializing themselves into fully composed MIME messages, since that's what Django's SMTP backend needs.)
* Just let boto3 deal with AWS credentials, etc.
* Message-ID is assigned by SES and returned from send call. Unclear what happens if we also provide a Message-ID header.
* SES doesn't have ESP templates or batch send. (Can ignore merge.)
* SES doesn't support delayed send_at.
[...]

PR #97, opened 2018-03-19
[WIP] Amazon SES integration

medmunds:
In-progress implementation of Amazon SES support (#54). This first commit implements all backend functionality other than templates.

For testing, pip can install directly from this feature branch:

```bash
pip install https://github.com/anymail/django-anymail/tarball/feature_amazon_ses#egg=django-anymail[amazon_ses]
```

(You may need to add `--upgrade` if you already have Anymail installed. The `#egg=django-anymail[amazon_ses]` part lets pip know to also install the extra dependencies for amazon_ses support in django-anymail—namely, boto3.)

To use the backend, specify `EMAIL_BACKEND = "anymail.backends.amazon_ses.EmailBackend"` in your Django settings. There are no specific Anymail settings required—[boto3 finds your AWS credentials](http://boto3.readthedocs.io/en/stable/guide/configuration.html) in its usual places.

TODO:
- Finish backend:
  - [x] Fix fragile tests (need better approach for comparing MIME parts)
  - [x] Implement `template_id` and `merge_data` using newish SES SendBulkTemplatedEmail API
  - [x] Resolve other open design/enhancement decisions (see TODO comments in source)
- Webhooks:
  - [x] Tracking webhooks
  - [x] Inbound webhook
  - ~Webhook [signature verification](https://docs.aws.amazon.com/sns/latest/dg/SendMessageToHttp.verify.signature.html)~ (holding out for something at least semi-official)
- [x] [Docs](http://anymail.readthedocs.io/en/feature_amazon_ses/esps/amazon_ses/)

### DIFF
--- a/anymail/backends/amazon_ses.py
+++ b/anymail/backends/amazon_ses.py
@@ -1,0 +1,179 @@
+from __future__ import absolute_import
+
+from .base import AnymailBaseBackend, BasePayload
+from ..exceptions import AnymailAPIError, AnymailImproperlyInstalled
+from ..message import AnymailRecipientStatus
+from ..utils import get_anymail_setting
+
+try:
+    import boto3
+    from botocore.exceptions import BotoCoreError, ClientError, ConnectionError
+except ImportError:
+    raise AnymailImproperlyInstalled(missing_package='boto3', backend='amazon_ses')
+
+
+# boto3 has several root exception classes; this is meant to cover all of them
+BOTO_BASE_ERRORS = (BotoCoreError, ClientError, ConnectionError)
+
+
+class EmailBackend(AnymailBaseBackend):
+    """
+    Amazon SES Email Backend (using boto3)
+    """
+
+    esp_name = "Amazon SES"
+
+    def __init__(self, **kwargs):
+        """Init options from Django settings"""
+        super(EmailBackend, self).__init__(**kwargs)
+        # AMAZON_SES_CLIENT_PARAMS is optional - boto3 can find credentials several other ways
+        self.client_params = get_anymail_setting("client_params", esp_name=self.esp_name,
+                                                 kwargs=kwargs, allow_bare=False, default={})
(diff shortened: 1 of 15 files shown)


anymail/backends/amazon_ses.py
```diff
@@ -0,0 +1,239 @@
...
+        if self.client:
+            return False  # already exists
+        try:
+            self.client = boto3.client("ses", **self.client_params)
+        except BOTO_BASE_ERRORS:
+            if not self.fail_silently:
+                raise
+
+    def close(self):
+        if self.client is None:
```
evgenyfadeev: looks like these couple of lines are unnecessary

medmunds: True, but I don't think they hurt anything, and they establish the pattern for closing the boto session if that ever makes sense... If (or when) there's a way to close a boto session, the comment immediately below would be replaced with something like `self.client.close()`. (Here's how that looks [in Requests-based backends](https://github.com/anymail/django-anymail/blob/feature_amazon_ses/anymail/backends/base_requests.py#L37-L46).)

medmunds:
Preliminary [Amazon SES docs](http://anymail.readthedocs.io/en/feature_amazon_ses/esps/amazon_ses/) are available.

I'm going to punt on [SNS signature verification](https://docs.aws.amazon.com/sns/latest/dg/SendMessageToHttp.verify.signature.html) until something at least semi-official is available from (or vaguely blessed by) Amazon. (The process is non-trivial enough that it doesn't belong in Anymail proper. Also, roll-your-own-crypto is, well, bad.)

That just leaves SendBulkTemplatedEmail and some real-world testing.

thongly:
This is amazing!  Thanks @medmunds .

We'll take it for a spin.  Will actually try to contribute too - just been slammed lately.

ducdetronquito:
It was the only missing feature in order to convince my company to switch to django-anymail !

Thanks a lot for your work :)

Issue #108, opened 2018-05-23
SendGrid smtp-id not always matching anymail_status


joshkersey:
ESP: SendGrid
django-anymail: 2.2
Python: 2.7
Django: 1.11.10

Just after we make the `msg = AnymailMessage(**kwargs)` call we store the data from `msg.anymail_status` in our database for tracking. The issue we're seeing is that the SendGrid webhook will often times return POST data with a `message_id` and `esp_event['smtp-id`] that do not match the values Anymail generated for us.

As an example:
We send a message and receive an smtp-id of `
<20180523175141.98405.51520@example.com>` in the `msg.anymail_status`.
When the webhook communicates updates to us we find that the `message_id` for the same message is `<YTJPMXaQSY-KzuEUmXrDWg@ismtpd0025p1las1.sendgrid.net>` which means that we are unable to identify which message this belongs to.

Here's the code that calls AnymailMessage:
```python
text_message = render_to_string(templates['text'], context)
html_message = render_to_string(templates['html'], context)
msg = AnymailMessage(
    subject=templates['subject'],
    body=text_message,
    to=[user.email],
)
msg.attach_alternative(html_message, "text/html")
msg.send()

# store the anymail status data to our database for tracking via the webhook
Email().create_from_anymail(msg.anymail_status, Email.INVITATION)
```

The `Email()` model holds the `create_from_anymail(anymail_status, event)` method where event is an optional custom tracker for the event triggering the email send (unrelated to anymail):
```python
@classmethod
def create_from_anymail(cls, anymail_status, event=None):
    for recipient in anymail_status.recipients:
        kwargs = {
            'to_email': recipient,
            'status': anymail_status.recipients[recipient].status,
            'message_id': anymail_status.recipients[recipient].message_id,
            'response': anymail_status.esp_response
        }

        if event:
            kwargs['event'] = event

        email = Email(**kwargs)
        email.save()
```

From what I can gather from the SendGrid webhook documentation the `smtp-id` should be returned for every webhook POST and is `a unique ID attached to the message by the originating system`.

I've also contacted SendGrid support but wanted to present the issue here in case anyone has seen something similar and may have ideas on how to get a consistent identifier from our system into SendGrid.

medmunds:
Anymail by default* generates a new message_id for each message it sends through SendGrid. It sets that in the message's Message-ID header, which is what SendGrid is supposed to pass back to webhook calls as the smtp-id field.

Two questions:

[...]

PR #109, opened 2018-05-24
Modified custom_arg message_id for compatibility with SendGrid webhook

joshkersey:
Based our discussion in #108.

According to SendGrid support, the `Message-ID` and `smtp-id` are both used by SendGrid to set their unique message ID. Based on the conversation with SendGrid support we should use a custom arg with a key named something other than `Message-ID` or `smtp-id`. 

Tried to set this up to ensure backwards compatibility for users who upgrade so that the `smtp-id` still gets referenced if the `anymail-id` is not present in the webhook data.

### DIFF
--- a/anymail/backends/sendgrid.py
+++ b/anymail/backends/sendgrid.py
@@ -110,7 +110,8 @@ class SendGridPayload(RequestsPayload):
     def set_anymail_id(self):
         """Ensure message has a known anymail_id for later event tracking"""
 
-        self.data.setdefault("custom_args", {})["anymail_id"] = str(uuid.uuid4())
+        self.message_id = str(uuid.uuid4())
+        self.data.setdefault("custom_args", {})["anymail_id"] = self.message_id
 
     def build_merge_data(self):
         """Set personalizations[...]['substitutions'] and data['sections']"""
--- a/anymail/backends/sendgrid.py
+++ b/anymail/backends/sendgrid.py
@@ -2,7 +2,6 @@ import uuid
 from email.utils import quote as rfc822_quote
 import warnings
 
-from django.core.mail import make_msgid
 from requests.structures import CaseInsensitiveDict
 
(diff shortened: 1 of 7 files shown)


anymail/backends/sendgrid.py
```diff
@@ -114,10 +114,8 @@ def ensure_message_id(self):
             self.data["headers"]["Message-ID"] = self.make_message_id()
         self.message_id = self.data["headers"]["Message-ID"]
 
-        # Workaround for missing message ID (smtp-id) in SendGrid engagement events
-        # (click and open tracking): because unique_args get merged into the raw event
-        # record, we can supply the 'smtp-id' field for any events missing it.
-        self.data.setdefault("custom_args", {})["smtp-id"] = self.message_id
+        # Set anymail-id to track our custom message ID
+        self.data.setdefault("custom_args", {})["anymail-id"] = self.message_id
```
medmunds: I'd maybe go with "anymail_id" (underscore), to parallel all other webhook data fields. ("smtp-id" is the outlier.)

anymail/webhooks/sendgrid.py
```diff
@@ -69,10 +69,12 @@ def esp_to_anymail_event(self, esp_event):
         else:
             metadata = {}
 
+        anymail_id = esp_event.get('anymail-id', None)
+        smtp_id = esp_event.get('smtp-id', None)
```
medmunds: This could be simplified to `message_id = esp_event.get('anymail_id', esp_event.get('smtp-id'))`.
(Use 'anymail_id' if present, else use 'smtp-id' for backwards compatibility.)

joshkersey:
Updated based on further discussion in #108. Removed a concern for `Message-ID` from the SendGrid and SendGrid V2 backends in favor of setting a UUID as `anymail_id` for tracking webhook notifications.

PR #126, opened 2018-10-05
also consider Content-ID when marking attachment as inline

### LINKED ISSUE #125: Mailgun backend loses attachments with non-ASCII filenames
* Anymail version: 4.2
* ESP: Mailgun
* Django: 1.11.8
* Python: 3.5

It took me a while to narrow down the circumstances, but it seems that if an attachment has a file name longer than 64 bytes, it's lost. It's simply not part of the e-mail. Neither in my mailbox, nor visible in the Mailgun logs (the attachment list is empty).

It's not just the string length of the file name. If I use special characters (like `æøå`, which we happen to use a lot in Denmark), the allowed string length is shortened.

Initially, I thought this was a Mailgun issue, but I've been in contact with their support, and they claim they don't have this limitation (they _do_ have a limit, but it's higher[1]). Also, if I send an e-mail with an attachment with a very long file name, based on [the `curl` example from their docs](https://documentation.mailgun.com/en/latest/api-sending.html#examples), the attachment is not lost.

If I send the e-mail using Django's built in SMTP backend, everything works fine with very long file names.

On the way, I thought special characters in general was the issue, but that's not the case:

 * The problem persists, even if I encode the file names with [`email.header.make_makeheader`](https://docs.python.org/3/library/email.header.html#email.header.make_header).
 * ASCII-only file names longer than 64 characters also fail.

I've been digging around the source, hoping to find a way to fix it myself, but didn't have any luck. If someone can point me in the right direction, I'd love to provide a fix myself.

[1] Quote from Mailgun support:

> When sending attachments, the size limit is 25mb, so you will also want to verify that your attachment falls under that limit. Regarding the file name length, there can be an argument made for the limit being as high as 255 and as low as 78 depending on the environment. Even if we err on the side of caution and treat 78 as the max allowed, your filename length at 64 should still be acceptable.

costela:
As explained in #125, this should bring `django-anymail` more inline with the behavior of many MUAs that optimistically consider an attachment as "inline" when it has a `Content-ID` but no `Content-Disposition`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,6 +33,10 @@ v4.3
 Features
 ~~~~~~~~
 
+*  Treat MIME attachments that have a *Content-ID* but no explicit *Content-Disposition*
+   header as inline, matching the behavior of many email clients. For maximum
+   compatibility, you should always set both (or use Anymail's inline helper functions).
+   (Thanks `@costela`_.)
 *  Add (undocumented) DEBUG_API_REQUESTS Anymail setting. When enabled, prints raw
    API request and response during send. Currently implemented only for Requests-based
    backends (all but Amazon SES and SparkPost). Because this can expose API keys and
--- a/anymail/utils.py
+++ b/anymail/utils.py
@@ -291,7 +291,8 @@ class Attachment(object):
                     self.content = attachment.as_string().encode(self.encoding)
             self.mimetype = attachment.get_content_type()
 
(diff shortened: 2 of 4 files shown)


costela:
Tests failing on `test_postmark_integration`? :thinking: 
Doesn't look like anything related to the change. Any ideas?

costela:
Yes, you're right. I got a bit carried away with the "fixes" line. Just removed it.

PR #131, opened 2018-10-26
webhooks/mailgun: mark temporary failures as DEFERRED

### LINKED ISSUE #130: [Proposal][Mailgun][webhooks] change event_types and reject_reason mappings
Currently, it seems impossible to differentiate temporary from permanent failures by analyzing `AnymailTrackingEvent` created by the mailgun webhook.
A delivery which had a temporary failure due to - for instance - a broken connection, will generate a webhook payload with (among other fields):
```json
{
  "event": "failed",
  "reason": "generic",
  "severity": "temporary"
}
```

This gets mapped to a `AnymailTrackingEvent` with `event_type=BOUNCED` and `reject_reason=BOUNCED`, which is hardly helpful. In particular, in the case of a failed connection, it feels actually misleading, IMHO, since there was no actual bounce from the receiving server.

Looking at `anymail/webhooks/mailgun.py`, I'd suggest doing something like the following in `esp_to_anymail_event()`:
```python
if event_type == EventType.BOUNCED and event_data['severity'] == 'temporary':
  event_type = EventType.DEFERRED
```
Also, changing the `reject_reasons` mapping from `generic → BOUNCED` to `generic → OTHER`. The rationale being: if Mailgun cannot interpret the error, we shouldn't assume it has any particular type.

Additionally, and mostly unrelated, we also see `"reason": "greylisted"` in our logs, which is currently not handled at all and might be mapped to `BOUNCED`?

Before working on a concrete PR, I thought I get some opinions on this. Thoughts?

costela:
Also remap "generic" failure reason to RejectReason.OTHER, instead  of
BOUNCE.

fixes #130

### DIFF
--- a/anymail/webhooks/mailgun.py
+++ b/anymail/webhooks/mailgun.py
@@ -87,16 +87,20 @@ class MailgunTrackingWebhookView(MailgunBaseWebhookView):
         # (these appear in webhook doc examples, but aren't actually documented anywhere)
         "bounce": RejectReason.BOUNCED,
         "suppress-bounce": RejectReason.BOUNCED,
-        "generic": RejectReason.BOUNCED,  # ??? appears to be used for any temporary failure?
+        "generic": RejectReason.OTHER,  # ??? appears to be used for any temporary failure?
+    }
+
+    severities = {
+        # Remap some event types based on "severity" payload field
+        (EventType.BOUNCED, 'temporary'): EventType.DEFERRED
     }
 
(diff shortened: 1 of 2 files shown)


costela:
Please note the minor style change when setting `event_type` for the first time. The `.get()` style (as opposed to `try/except`) was already being used for `reject_reason` and makes the code IMHO slightly more homogenous. Let me know if the `try/except` style is preferred.

medmunds:
Looks great. Thanks for this.

PR #156, opened 2019-07-29
Add template support for Mailgun

### LINKED ISSUE #155: Mailgun Template Support
The docs claim that
> Mailgun does not offer ESP stored templates, so Anymail’s template_id message attribute is not supported with the Mailgun backend.

but that does not appear to be true: https://documentation.mailgun.com/en/latest/api-templates.html

Am I mistaking either Anymail's or Mailgun's docs?

anstosa:
Fixes #155

### DIFF
--- a/tests/test_mailgun_backend.py
+++ b/tests/test_mailgun_backend.py
@@ -448,7 +448,11 @@ class MailgunBackendAnymailFeatureTests(MailgunBackendMockAPITestCase):
         self.assertEqual(data['o:tracking-opens'], 'no')
         self.assertEqual(data['o:tracking-clicks'], 'yes')
 
-    # template_id: Mailgun doesn't support stored templates
+    def test_template_id(self):
+        self.message.template_id = "welcome_template"
+        self.message.send()
+        data = self.get_api_call_json()
+        self.assertEqual(data['template'], "welcome_template")
 
     def test_merge_data(self):
         self.message.to = ['alice@example.com', 'Bob <bob@example.com>']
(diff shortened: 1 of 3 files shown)


tests/test_mailgun_backend.py
```diff
@@ -448,7 +448,11 @@ def test_tracking(self):
         self.assertEqual(data['o:tracking-opens'], 'no')
         self.assertEqual(data['o:tracking-clicks'], 'yes')
 
-    # template_id: Mailgun doesn't support stored templates
+    def test_template_id(self):
+        self.message.template_id = "welcome_template"
+        self.message.send()
+        data = self.get_api_call_json()
```
medmunds: Changing this to `self.get_api_call_data()` should make the tests pass. (Mailgun's API uses form field encoding, rather than JSON.)

anstosa: Whoops, thanks! Updated.

ryanovas:
Is there a planned time at which this will be available via pip? Right now I am only able to see 6.1 which doesn't support mailgun templates.

medmunds:
@ryanovas this will appear in Anymail v7.0, which I'm hoping to have time to finish up and release over the weekend or early next week. Sorry for the delay; it took [some research](https://github.com/anymail/django-anymail/blob/feature_mailgun_templates/anymail/backends/mailgun.py#L125-L170) to figure out how to make Mailgun's new templates work with Anymail's merge data without breaking the other Mailgun template format or metadata.

If you're in a rush, see [this comment](https://github.com/anymail/django-anymail/issues/155#issuecomment-522725612) in the original issue for instructions on using Mailgun's new templates with older versions of Anymail. (Anymail has an `esp_extra` override specifically so people don't have to wait on an Anymail update to use newly-released ESP features.)

medmunds:
Released in Anymail v7.0

PR #158, opened 2019-08-28
Improvements to Sendinblue Backend

Thorbenl:
- Multiple tags can be send now (regardless of template or not)
- When sending a mail using a template, display name is supported on 'to', 'bcc', 'cc' and 'replyTo'. Also supports overriding 'from_email' and 'subject'
- Removed use of deprecated SendinBlue API endpoint

### DIFF
--- a/anymail/backends/sendinblue.py
+++ b/anymail/backends/sendinblue.py
@@ -1,5 +1,4 @@
 from requests.structures import CaseInsensitiveDict
-from six.moves.urllib.parse import quote
 
 from .base_requests import AnymailRequestsBackend, RequestsPayload
 from ..exceptions import AnymailRequestsAPIError
@@ -67,7 +66,6 @@ class SendinBluePayload(RequestsPayload):
 
     def __init__(self, message, defaults, backend, *args, **kwargs):
         self.all_recipients = []  # used for backend.parse_recipient_status
-        self.template_id = None
 
         http_headers = kwargs.pop('headers', {})
         http_headers['api-key'] = backend.api_key
@@ -76,10 +74,7 @@ class SendinBluePayload(RequestsPayload):
         super(SendinBluePayload, self).__init__(message, defaults, backend, headers=http_headers, *args, **kwargs)
 
     def get_api_endpoint(self):
-        if self.template_id:
-            return "smtp/templates/%s/send" % quote(str(self.template_id), safe='')
-        else:
-            return "smtp/email"
+        return "smtp/email"
 
     def init_payload(self):
         self.data = {  # becomes json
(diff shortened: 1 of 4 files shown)


medmunds:
@Thorbenl Thanks for this! Better functionality and simpler code too—gotta like it.

medmunds:
Oh, hmm... apparently you can't send attachments with templates using the new API. [You could](https://github.com/anymail/django-anymail/blob/v6.1/tests/test_sendinblue_integration.py#L77-L93) with the old (now deprecated) one.

```
{
  "code": "invalid_parameter", 
  "message": "Please don't pass attachment content & templateId in same request, instead use attachment url only"
}
```

(Well, you *can* still send attachments, but only if you first upload the attachment content somewhere and pass Sendinblue a public URL to it, which... um... well that's just weird. And way outside the scope of Anymail. I'm struggling to even conceive of a backend architecture where it's somehow easier for them to download attachment content separately rather than just be handed the data at the outset.)

I guess this is a new Sendinblue limitation to document, and counts as a breaking change for the next release. I suspect attachments with templates are relatively rare, and having recipient display names work properly with templates outweighs the loss of attachments.

(BTW, was any of this ever documented or announced anywhere? I can't find anything useful about [template](https://blog.sendinblue.com/?s=template) or [tags](https://blog.sendinblue.com/?s=tags) changes in their blog, and their [changelog](https://developers.sendinblue.com/docs/change-log) has nothing about tags, and only a cryptic reference to "Old Template Language (i.e. Version 1)"—really nothing about about updates to the transactional send call.)

Thorbenl:
@medmunds Thanks for approving and merging! :) This wasn't really announced, as far as I can see  sadly. I found out by poking around with Postman. We are currently integrating Sendinblue into our application, and did some prepwork to find out what the api supports and what not. Their docs are pretty descriptive in some sense, but lack in other :) 

Yeah, there wasnt any docs about how to update the send call or when it was changed.

medmunds:
This has [broken the integration tests](https://travis-ci.org/anymail/django-anymail/builds/578141520), so I'll need to update those (and the docs as noted above), before it can be released. I'm going to try to finish a Mailgun template update that's also been on the list for a while, and release a new version in the next couple of days.

medmunds:
@Thorbenl I've been doing some additional testing directly to the Sendinblue API (not going through Anymail). It seems like the v3/smtp/email has some serious limitations when `templateId` is used:

1. Although including a display name in `to` or `cc` doesn't cause an error, the `name` fields seem to be silently ignored and don't appear in the sent message
2. `replyTo` seems to be silently ignored (but works with the deprecated API)
3. `attachment` content causes an error (but works with the deprecated API, as noted earlier)
4. `sender` (from_email) and `subject` overrides are silently ignored—the template version is always used

From what I can tell, the only new functionality that actually works as documented is the new `tags` array.

Since items 2 and 3 worked correctly with the deprecated v3/smtp/templates/*templateId*/send API, and display names and template overrides *don't* seem to work any better with the new API than the old one (which didn't support them at all), I'm not sure it really makes sense to release this change in Anymail.

Oddly enough, the deprecated template send API seems to support the new `tags` parameter.

Are you seeing different behavior? (Do display names actually make it into the delivered email when you send with a template? Subject overrides?) If things are working for you, I'd like to figure out what's different about my test case before releasing this, and include relevant info in the docs.

But if you're seeing the same template send limitations as I am, then I think the right thing to do is probably:
* Revert the merge into master and reopen this PR
* You (or I) edit the PR to keep the new `tags` array support, but put template handling back the way it was using the deprecated API
* Re-merge after that
* Maybe open an issue with Sendinblue support? (Particularly if you were planning to use any of this functionality in your integration.)

What do you think?

Thorbenl:
Hey @medmunds. I am sorry if you had some trouble. So, I have verified that overriding subject/from_email works, as well as setting replyTo and custom to/cc/replyTo names. While attachments does indeed have the limitation that you can only specify a URL and not the file contents directly. I will sent you the curl payload i used to verify:

```json
curl --request POST \
  --url https://api.sendinblue.com/v3/smtp/email \
  --header 'accept: application/json' \
  --header 'api-key: YOUR_API_KEY_HERE' \
  --header 'content-type: application/json' \
  --data '{"sender":{"name":"SenderName","email":"sender@example.com"},"to":[{"email":"to@example.com","name":"ToName"}],"cc":[{"email":"cc@example.com","name":"CCName"}],"subject":"SubjectOverride","replyTo":{"email":"replyto@example.com","name":"ReplytoName"},"templateId":1}'
```
Dont foget to use your API key

medmunds:
@Thorbenl figured it out: *if* (and only if) your template contains old-style `%PARAM%` substitutions, then none of the new features work. More details in #161.

I'm guessing that, internally, Sendinblue has coded the v3/smtp/send API to check whether the template content contains any old-style `%PARAM%`s, and if so, redirect to the deprecated v3/smtp/templates/templateId/send call, dropping any payload data that doesn't have a direct mapping to the old API (like display names).

If you don't have any `%PARAM%`s, then your old-style template is indistinguishable from a new-style template, and works fine on the new code path. (Their web template editor seems to use the same logic for deciding whether to show the "upgrade" link.)

medmunds:
Released in Anymail v7.0

PR #159, opened 2019-08-30
Mailgun: make merge_data work with stored handlebars templates

### LINKED ISSUE #155: Mailgun Template Support

medmunds:
Mailgun has two different template mechanisms and two different ways
of providing substitution variables to them. Update Anymail's
normalized merge_data handling to work with either (while preserving
existing batch send and metadata capabilities that also use Mailgun's
custom data and recipient variables parameters).

Completes work started by @anstosa in #156.
Closes #155.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -25,6 +25,19 @@ Release history
 ^^^^^^^^^^^^^^^
     ..  This extra heading level keeps the ToC from becoming unmanageably long
 
+vNext
+-----
+
+*Not yet released*
+
+Features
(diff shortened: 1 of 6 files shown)


medmunds:
Released in Anymail v7.0

PR #160, opened 2019-09-03
Mailgun: disable RFC 5758 filename workaround when urllib3 is fixed

### LINKED ISSUE #157: Mailgun: disable workaround for Unicode attachment filenames if fixed urllib3 in use
Anymail v4.3 added a workaround for an urllib3 issue that caused Mailgun problems when trying to send attachments with non-ASCII filenames. (See #125)

The underlying [urllib3 issue](https://github.com/urllib3/urllib3/issues/303) was [fixed](https://github.com/urllib3/urllib3/pull/1492) in urllib3 v1.25 (2019-04-22).

Anymail's Mailgun backend should feature-detect that urllib3 fix, and disable the filename workaround in `MailgunPayload.get_request_params` when it's not required. (The workaround ends up being a no-op with a newer urllib3, but there's a bit of needless overhead along the way.)

medmunds:
urllib3 v1.25 fixes non-ASCII filenames in multipart form data to be
RFC 5758 compliant by default, so our earlier workaround is no longer
needed. Disable the workaround if we detect that Requests is using a
fixed version of urllib3.

Closes #157

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,6 +52,9 @@ matrix:
     - { env: TOXENV=django22-py37-none, python: 3.7 }
     - { env: TOXENV=django22-py37-amazon_ses, python: 3.7 }
     - { env: TOXENV=django22-py37-sparkpost, python: 3.7 }
+    # Test some specific older package versions
+    - { env: TOXENV=django111-py27-all-old_urllib3, python: 3.7 }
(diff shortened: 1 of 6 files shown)


medmunds:
Released in Anymail v7.0

PR #166, opened 2019-09-17
Add Postal support

tiltec:
This is a very bare-bone implementation for the self-hosted [Postal](https://github.com/atech/postal) platform.

About Postal:

> Postal is a complete and fully featured mail server for use by websites & web servers. Think Sendgrid, Mailgun or Postmark but open source and ready for you to run on your own servers.

We use this draft implementation since a few days to handle emails for our [karrot.world](https://karrot.world) web app. Before we used SparkPost via anymail. I had good experiences with anymail so far, that's why I'm keen on using it.
I'll improve this implementation during the next weeks as I find time. I'm happy about early feedback!

To do:

- [x] figure out how to do webhook authentication with Postal
- [ ] support track_clicks and track_opens
- [x] add backend tests
- [x] add webhook tests
- [x] add docs
- [x] integration tests?


medmunds:
This is great, thanks! I'm out of town right now, so just took a quick look at the code, but it looks good so far. Let me know if there are any specific questions I can answer.

As far as webhook auth, if Postal doesn't already have some means of signing webhook payloads, using Anymail's built-in http basic auth should be sufficient.

My memory is that aTech no longer offers a hosted Postal service, so we probably won't be able to include this in Anymail's live CI tests (unless you know someone running Postal who could set up an Anymail test account). But I suspect the Postal API isn't changing much, so regular live API testing might be less important than with other ESPs.

tiltec:
Thanks for having a look! Implementation is working fine so far, I'll see when I can get around to bring the PR into a mergable state.

> As far as webhook auth, if Postal doesn't already have some means of signing webhook payloads, using Anymail's built-in http basic auth should be sufficient.

I wasn't able to get http basic auth working with Postal. I didn't investigate more, but it clearly didn't seem to be designed for this use case: the webhook URL validation doesn't allow the `@` character.

> My memory is that aTech no longer offers a hosted Postal service, so we probably won't be able to include this in Anymail's live CI tests (unless you know someone running Postal who could set up an Anymail test account). But I suspect the Postal API isn't changing much, so regular live API testing might be less important than with other ESPs.

As Postal is open-source, we could spin up our own testing instance in Travis (e.g. via docker image). But as you said, the API isn't changing much, so live tests seem safe to be left out for now.

medmunds:
> I wasn't able to get http basic auth working with Postal. I didn't investigate more, but it clearly didn't seem to be designed for this use case: the webhook URL validation doesn't allow the @ character.

Ah. The nice thing about using HTTP basic auth for webhook authentication is that it tends to just work with software that hasn't put any special effort into webhook auth. Unless, you know, that software [rolls its own regexp](https://github.com/SynergiTech/postal/blob/634d3af59d392e84bdcd2b843b427f3a79d3ff4c/app/models/webhook.rb#L31) to (not quite accurately) validate webhook urls.

However, it sounds like Postal [signs the webhook body](https://github.com/atech/postal/issues/432#issuecomment-357457313), apparently [using SHA-1](https://github.com/atech/encrypto-signo/blob/1de9ec165b864f613cf359fb4ebc18a04273ec82/lib/encrypto_signo.rb#L28) with /opt/postal/config/signing.key, and passes the base64-encoded signature in the `X_POSTAL_SIGNATURE` header. So I suppose you could verify that instead. (And include a docs warning box about SHA-1 security.)

medmunds:
> So far, cryptography is not a dependency of anymail - do you think it's fine if I add it in setup.py as extras_require for Postal?

Yes, perfect.

anymail/webhooks/postal.py
```diff
@@ -0,0 +1,169 @@
...
+        if message_to is not None:
+            try:
+                recipient = parse_single_address(message_to).addr_spec
+            except AnymailInvalidAddress:
+                pass
+
+        if message.get("direction") == "incoming":
+            # this event is about an incoming email being delivered to Anymail
+            # we can safely ignore this
+            print("ignoring")
```
medmunds: You could raise an `AnymailConfigurationError` here, to help folks who have accidentally swapped Anymail webhook urls. ([Example](https://github.com/anymail/django-anymail/blob/v7.0/anymail/webhooks/postmark.py#L79-L82) from Postmark webhooks.)

setup.py
```diff
@@ -54,6 +54,7 @@ def long_description_from_readme(rst):
         "postmark": [],
         "sendgrid": [],
         "sendinblue": [],
+        "postal": ["cryptography"],
```
medmunds: `cryptography` will also need to be in `tests_require` below.

medmunds:
For adding cryptography, you'll also need to update tox.ini: extend the [testenv config](https://github.com/anymail/django-anymail/blob/v7.0/tox.ini#L32-L39) to install cryptography when the `postal` "factor" is being tested, and add the `postal` factor to the [partial installation](https://github.com/anymail/django-anymail/blob/v7.0/tox.ini#L17) tests. And then add a line to .travis.yml to [kick off](https://github.com/anymail/django-anymail/blob/v7.0/.travis.yml#L51-L54) a `postal`-only extras installation test.

anymail/webhooks/postal.py
tiltec: I noticed this in other implementations, but this case is a bit different. This is not the inbound email, but a tracking event _about_ the inbound email. So it's the right webhook, but it's not interesting to anymail users (and there's no `EventType` for it anyway).

medmunds: Oh, that's weird. This is the first ESP I've encountered that generates tracking events for inbound messages. (I'm not even really sure what it means to track a message you didn't send.) But yeah, probably not something we need to normalize for Anymail.

anymail/webhooks/postal.py
```diff
@@ -0,0 +1,169 @@
...
+    EventType,
+)
+
+try:
+    from cryptography.exceptions import InvalidSignature
+    from cryptography.hazmat.backends import default_backend
+    from cryptography.hazmat.primitives import serialization, hashes
+    from cryptography.hazmat.primitives.asymmetric import padding
+except ImportError:
+    raise AnymailImproperlyInstalled(missing_package='cryptography', backend='postal')
```
medmunds: You'll need to use `_LazyError` here, or a missing cryptography will cause errors with other ESPs. See the amazon_ses webhooks for an example.

tiltec:
I probably can't spend much time on this in the next weeks, so I'd be interested in your preferred way forward. I'm happy to keep this as WIP PR, but I'd also spend some hours to get it into a mergeable state. What do you think?

medmunds:
I'm happy to add docs and tests, but it'll probably be a week or two before I'll be able to get to it. Let's keep the PR as WIP for now. I'll leave a comment before I start any work, just in case you get back to it before I do.

tiltec:
Coming back to this, I just merged in the latest changes but couldn't make the CI run:

image

It says "Branch master not included per configuration". Is this on purpose?

medmunds:
Sorry about that. I renamed the default git branch to "main" a few months back. (I was under the impression GitHub would be automatically patching up open PRs, but apparently that isn't working yet.) You'll need to rebase the PR onto upstream/main. (And I should probably delete the now-obsolete master branch.)

The other major change this year is dropping Python 2 and older Django versions, which ended up removing a *lot* of code. I don't exactly have a checklist, but the series of commit comments in #193 might be helpful (maybe).

tiltec:
Thanks! I changed the base branch of this PR and went through the commits in #193, seems fine! The PR still didn't build in Travis, but it seems due to it being a "Draft". I removed the draft status and pushed again, but no success yet. I'll wait a bit and try again soon.

My plan is to get this PR eventually merged into master. It has been working fine in production for over a year and possibly it might be useful for others as well. Postal itself seems to be barely maintained, but it's the only open-source transactional email software I'm aware of, which makes it support-worthy to me.

I guess I would continue with adding some docs, then possibly look into tests. I'd leave out click- and open tracking because I don't use it myself. Does that sound alright?

medmunds:
Thank you! This is looking great.

It looks like the builds are running (and [this PR built successfully](https://travis-ci.org/github/anymail/django-anymail/pull_requests)), but the status isn't getting reported back here to GitHub. I suspect that's related to ongoing changes in TravisCI's support for open source projects under their new owners. I'm going to look into a new CI solution.

Yes, we'd want docs and tests before merging the PR. For basic sending tests, it looks like Postal's API is similar to Postmark's, so you might be able to borrow from [test_postmark_backend.py](https://github.com/anymail/django-anymail/blob/v8.1/tests/test_postmark_backend.py). I think it would be important to test the webhooks as well, particularly signature validation.

For click and open tracking, Anymail just tries to allow communicating whether those features should be enabled/disabled on an individual message. It looks like [Postal supports disabling both](https://github.com/postalhq/postal/wiki/Click-&-Open-Tracking#disable-tracking-on-a-per-e-mail-basis) simultaneously by adding a message header. We could try to get fancy and add that header if the user sets _both_ `track_clicks` and `track_opens` to False (and issue an unsupported feature error for any other combination). But I think it's fine to just leave them both unimplemented for now (which defaults to unsupported feature).

It looks like Postal's [send/message API](https://krystal.github.io/postal-api/controllers/send/message) supports two other features that Anymail tries to normalize: attachments and (envelope) sender. It would be nice to support them if possible, but we could certainly leave that for later too.

tests/test_postal_backend.py
```diff
@@ -0,0 +1,473 @@
...
+        """If the send succeeds, but a non-JSON API response, should raise an API exception"""
+        mock_response = self.set_mock_response(status_code=200,
+                                               raw=b"yikes, this isn't a real response")
+        with self.assertRaises(AnymailAPIError):
+            self.message.send()
+        self.assertIsNone(self.message.anymail_status.status)
+        self.assertIsNone(self.message.anymail_status.message_id)
+        self.assertEqual(self.message.anymail_status.recipients, {})
+        # TODO esp_response is None because post_to_esp calls raise_for_status which tries to deserialize the response
+        # self.assertEqual(self.message.anymail_status.esp_response, mock_response)
```
tiltec, line 435: This is an interesting one, I wonder what to do here.

In `raise_for_status`, called by `post_to_esp`, I need to deserialize the JSON response to figure out if the request was successful, because Postal doesn't really use HTTP status codes. If the response is unparsable, this would raise an error, not allowing `AnymailBaseBackend._send` to set `esp_response`:
https://github.com/anymail/django-anymail/blob/a7ea862ff147b8b9ce18786a194add199c1a2630/anymail/backends/base.py#L123-L125

I guess I could catch the deserialization error in `raise_for_status` and _not_ raise an error in this case. It will be raised later in `parse_recipient_status`.

tiltec:
Thank you for the helpful hints! I added backend tests, as well as support for attachments and envelope sender. 

I had to leave out support for inline images, since this seems only supported with the [raw RFC2882 send endpoint](https://krystal.github.io/postal-api/controllers/send/raw.html).

For now, I also didn't do anything about supporting tracking because my Postal instance isn't set up to support it yet. Might tackle this at the very end, or leave it out.

Next, I would work on webhook tests.

anymail/webhooks/postal.py
```diff
@@ -0,0 +1,175 @@
...
+
+            if "bounce" in event:
+                return EventType.BOUNCED
+
+            if "url" in event:
+                return EventType.CLICKED
+
+            return EventType.UNKNOWN
+
+        event_type = get_event_type(payload)
```
medmunds, line 109: Rather than defining and calling a local function, could I suggest just using `if ... elif ... elif ... else ... ` here to assign `event_type`?

anymail/webhooks/postal.py
```diff
@@ -0,0 +1,175 @@
...
+        if message_to is not None:
+            try:
+                recipient = parse_single_address(message_to).addr_spec
+            except AnymailInvalidAddress:
+                pass
+
+        if message.get("direction") == "incoming":
+            # This is a tracking event about inbound email, so it's not interesting to anymail.
+            # We can safely ignore this.
+            return []
```
medmunds, line 129: For other ESPs, Anymail raises a ConfigurationError ([example](https://github.com/anymail/django-anymail/blob/v8.1/anymail/webhooks/postmark.py#L79-L82)) in this case, because swapping webhook urls is a pretty common configuration mistake. (Assuming Postal lets you set up different webhooks for different event types; I can't tell from their docs.)

anymail/webhooks/postal.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+class PostalInboundWebhookView(PostalBaseWebhookView):
+    """Handler for Postal inbound relay webhook"""
+
+    signal = inbound
+
+    def parse_events(self, request):
+        esp_event = json.loads(request.body.decode("utf-8"))
+
```
medmunds, line 161: Similar to earlier comment, if there's a field you can check to make sure this actually *is* an inbound event, and issue a ConfigurationError if not, that's helpful. (If not, don't worry about it.)

tests/test_postal_backend.py
```diff
@@ -0,0 +1,473 @@
...
+        self.assertEqual(attachments[0]["name"], "test.txt")
+        self.assertEqual(attachments[0]["content_type"], "text/plain")
+        self.assertEqual(decode_att(attachments[0]["data"]).decode('ascii'), text_content)
+
+        self.assertEqual(attachments[1]["content_type"], "image/png")  # inferred from filename
+        self.assertEqual(attachments[1]["name"], "test.png")
+        self.assertEqual(decode_att(attachments[1]["data"]), png_content)
+
+        self.assertEqual(attachments[2]["content_type"], "application/pdf")
+        self.assertEqual(attachments[2]["name"], "")  # none
```
medmunds, line 174: BTW, it's possible that the real Postal API will actually issue an `AttachmentMissingName` error in this case (based on [the docs](https://krystal.github.io/postal-api/controllers/send/message.html#:~:text=AttachmentMissingName)). There are some other ESPs with the same limitation, and Anymail usually just documents that as an ESP-specific quirk.

tests/test_postal_backend.py
medmunds: See my comment in the backend. I think the best fix is for the postal backend to *not* implement `raise_for_status`, and to move the error checking into `parse_recipient_status`.

medmunds (changes requested):
Again, this is looking great. I really appreciate the tests.

I made a few specific suggestions (and added some questions) inline.

medmunds:
@tiltec I just migrated CI off Travis to GitHub Actions, so you should see a test run here after you next rebase and push.

Apologies for the merge conflict. .travis.yml no longer exists (just delete it). You'll instead need to add your Postal-only test run to [.github/workflows/test.yml](https://github.com/anymail/django-anymail/blob/main/.github/workflows/test.yml#L54-L56):

```patch
           # Install without optional extras (don't need to cover entire matrix)
           - { tox: django31-py37-none, python: 3.7 }
           - { tox: django31-py37-amazon_ses, python: 3.7 }
+          - { tox: django31-py37-postal, python: 3.7 }
           # Test some specific older package versions
           - { tox: django22-py37-all-old_urllib3, python: 3.7 }
 ```

(Or if you'd prefer, I'm happy to rebase and update the PR for you. Just let me know.)

File: anymail/webhooks/postal.py (hunk 6 above)
Comment (tiltec) [reply anymail/webhooks/postal.py:138]: done! 

I also tried to clarify the purpose of the code above, does it make sense to you?

```
            # Let's ignore tracking events about an inbound emails.
            # This happens when an inbound email could not be forwarded.
            # The email didn't originate from Anymail, so the user can't do much about it.
            # It is part of normal Postal operation, not a configuration error.
```

File: anymail/webhooks/postal.py (hunk 7 above)
Comment (tiltec) [reply anymail/webhooks/postal.py:170]: Done!

File: tests/test_postal_backend.py (hunk 4 above)
Comment (tiltec) [reply tests/test_postal_backend.py]: Done!

File: anymail/webhooks/postal.py (hunk 5 above)
Comment (tiltec) [reply anymail/webhooks/postal.py]: Done!

**Comment by tiltec** [comment]
Ah nice! I added your suggestion to `test.yml`, let's see if a test run shows up here.

Other than this, I'm slowly progressing with testing and refinements. Backend, inbound and webhook tests should be mostly there, but I'll probably have another look at it in a few days.

In the meantime, I'd start writing some docs and think about integration testing.

File: runtests.py [hunk 9]
```diff
@@ -20,6 +20,7 @@ def setup_and_run_tests(test_labels=None):
 
     tags = envlist('ANYMAIL_ONLY_TEST')
     exclude_tags = envlist('ANYMAIL_SKIP_TESTS')
+    print('exclude', exclude_tags)
```
medmunds, line 23: Oops, looks like some debugging code crept into your commit

medmunds (changes requested):

medmunds:
Sorry, I missed a "pull_requests" trigger in tests.yml. If you rebase again, it will (hopefully) run CI on this PR.

The webhook tests look great. I like what you've done with the Django test client that handles the signing—keeps everything really clean. I think for testing purposes we could probably get away with a static signing key (rather than creating a new one on every test run), but that would likely cause GitHub to send all kinds of panicked notifications about exposed private keys in the source code, so probably better the way you have it.

For the integration tests, my suggestion is to write them (borrow from, e.g., the Postmark tests), temporarily change the from and recipient addresses to your own (real, personal) email addresses, and run them use your own Postal instance. Check that the messages you receive contain the expected content—particularly around attachments and headers (the "all options" test is helpful for this). My experience with *every* other ESP so far is that these tests have exposed cases where their docs don't quite match their actual behavior—though maybe not an issue with Postal since you've been working with it live along the way.

Once you've done that, change the addresses back to example.com, and add the integration tests to the PR. They'll get skipped in GitHub CI (because the environment variables won't be set). But they'll be ready to go for anyone debugging locally, or if we later want to experiment with firing up a Postal instance during CI (which definitely doesn't need to be part of this PR).

tests/test_postal_backend.py
tiltec: Done!

runtests.py
tiltec: Indeed, thanks for catching it!

medmunds:
@tiltec this is looking really solid. I'm good with merging it whenever you're ready.

medmunds (approved):

tiltec:
Ok, arriving back at this. Sorry for the silence, got busy with kids!

I merged in latest changes, ran the tests and pushed. @medmunds could you approve the workflow to run all the tests?

From my perspective, it seems fine to merge. We are definitely going to keep using Postal in production, so I would be happy to provide some support if others are going to use Postal with anymail in future.

medmunds:
@tiltec Thanks for all the effort you put into this. It's a really nice addition to the project.

medmunds:
Released in Anymail v8.4

PR #177, opened 2020-02-27
SendinBlue: Attachments now work with templates

sebashwa:
Attachments now work with templates as long as you are using the new
template language.

I just tried it with `message.attach_file("/path/to/file.ext")` using a template and it works like a charm. This is very fortunate.

See also: https://developers.sendinblue.com/changelog/send-attachment-with-template-through-apiv3

Thanks for this awesome library!

### DIFF
--- a/docs/esps/sendinblue.rst
+++ b/docs/esps/sendinblue.rst
@@ -130,12 +130,9 @@ SendinBlue can handle.
   Anymail has no way to communicate an attachment's desired content-type
   to the SendinBlue API if the name is not set correctly.
 
-**No attachments with templates**
-  If you are sending using a SendinBlue template, their API doesn't support ordinary
-  file attachments. Attempting to send an attachment with a template will result in the
-  SendinBlue API error message, "Please don't pass attachment content & templateId in same
-  request, instead use attachment url only."
-  See the :ref:`templates <sendinblue-templates>` section below.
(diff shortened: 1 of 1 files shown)


medmunds:
Thanks for catching the change. Appreciate the update.

Anymail only supports SendinBlue's *new* template language anyway, so we could probably just remove any mention of template attachments from the limitations section.

docs/esps/sendinblue.rst
```diff
@@ -130,12 +130,9 @@ SendinBlue can handle.
...
 
-**No attachments with templates**
-  If you are sending using a SendinBlue template, their API doesn't support ordinary
-  file attachments. Attempting to send an attachment with a template will result in the
-  SendinBlue API error message, "Please don't pass attachment content & templateId in same
-  request, instead use attachment url only."
-  See the :ref:`templates <sendinblue-templates>` section below.
+**Attachments using templates only with new template language**
+  If you are sending an email using a SendinBlue template,
+  you have to use the new template language in order to add file attachments.
```
medmunds: (We could probably just omit this: Anymail v7 or later requires the new template language no matter what, even without attachments.)

sebashwa: Agreed and updated 👍

medmunds:
Thanks @sebashwa

Issue #179, opened 2020-03-12
After upgrade to 7.x - it is impossible send one message with multiple recipients


vgrebenschikov:
* django-anymail[sendgrid]==7.0.0
* EMAIL_BACKEND = 'anymail.backends.sendgrid.EmailBackend'
* Django==1.11.20, requests==2.21.0, python 3.7.4
* AssertionError at sengrid.py:137 in build_merge_data
```
assert len(personalization["to"]) == 1
```

Looks like problem was introduced in [d2d568b6d3](https://github.com/anymail/django-anymail/commit/d2d568b6d33acf0a0893dd107d96af02f5993946)
(@medmunds - can you advice was it intentional?)

before - assert on multiple personalizations:
```
assert len(self.data["personalizations"]) == 1
```
after - assert on multiple recipients in one mail:
```
assert len(personalization["to"]) == 1
```
Which completely broke my case when multiple recipients are specified in to_addresses (see below)

Stacktrace:
> File ".../django/core/mail/message.py", line 348, in send
>      return self.get_connection(fail_silently).send_messages([self])
> File ".../anymail/backends/base.py", line 95, in send_messages
>      sent = self._send(message)
> File ".../anymail/backends/base_requests.py", line 60, in _send
>      return super(AnymailRequestsBackend, self)._send(message)
> File ".../anymail/backends/base.py", line 125, in _send
>      response = self.post_to_esp(payload, message)
> File ".../anymail/backends/base_requests.py", line 71, in post_to_esp
>      params = payload.get_request_params(self.api_url)
> File ".../anymail/backends/base_requests.py", line 166, in get_request_params
>      data=self.serialize_data(),
> File ".../anymail/backends/sendgrid.py", line 104, in serialize_data
>      self.build_merge_data()
> File ".../anymail/backends/sendgrid.py", line 137, in build_merge_data
>      assert len(personalization["to"]) == 1
> AssertionError

```
from django.core.mail import EmailMultiAlternatives

def send_mail(
    template_id, subject='', body='', body_html='', from_address=None, to_addresses=None,
    context=None,
):
    msg = EmailMultiAlternatives(subject, body, from_address, to_addresses)

    if template_id:
        msg.merge_global_data = context
        msg.template_id = template_id

    if body_html:
        msg.attach_alternative(body_html, 'text/html')

    try:
        msg.send()
    except AnymailError:
        logger.exception("Failed to send message with Anymail")
```

medmunds:
[...]

PR #180, opened 2020-03-12
treat message with 'merge_global_data' attribute as batch to force personalizations expansion

jazz-jack:
don't know if this correct for all cases, but it works when we need same data for multiple recipients
see #179

### DIFF
--- a/anymail/backends/base.py
+++ b/anymail/backends/base.py
@@ -252,7 +252,7 @@ class BasePayload(object):
 
     # If any of these attrs are set on a message, treat the message
     # as a batch send (separate message for each `to` recipient):
-    batch_attrs = ('merge_data', 'merge_metadata')
+    batch_attrs = ('merge_data', 'merge_global_data', 'merge_metadata')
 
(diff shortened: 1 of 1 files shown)


jazz-jack:
the idea is to run ```expand_personalizations_for_batch``` in case if substitutions specified with ```merge_global_data``` as it may contain multiple ```to``` addresses

medmunds:
Thanks for this, but I don't believe it's the correct fix for #179. Setting only `merge_global_data` without `merge_data` is *not* supposed to put Anymail into batch send mode. (There's even a specific test for this, `test_merge_global_data_does_not_imply_batch`, which is why this PR is failing builds.)

I think we can just relax the assertion that's preventing SendGrid from working properly with only `merge_global_data` and multiple recipients. See [my comment](https://github.com/anymail/django-anymail/issues/179#issuecomment-599114376) in #179.

jazz-jack:
> Thanks for this, but I don't believe it's the correct fix for #179. 
sure, that was only a hint on the issue. thanks, Mike!

PR #182, opened 2020-04-13
Postmark: Send a templated email with a single recipient

jc-ee:
Hi,

This PR fixes a bug in the Postmark backend.
Attempting to send a templated email with a single recipient fails, as a `TemplateModel` is not being set. This is required for the [/email/withTemplate/](https://postmarkapp.com/developer/api/templates-api#email-with-template) endpoint.

This was causing the send to fail with a `Postmark API response 422 (Unprocessable Entity)`

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,6 +33,8 @@ vNext
 Fixes
 ~~~~~
 
+* **Postmark:** Fix API error when sending with template to single recipient.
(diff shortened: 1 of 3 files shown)


medmunds:
Thanks! Good catch, and I appreciate the fix and test.

While I'm looking at this, I'm going to add some defensive code to try to avoid similar problems in the future.

medmunds:
Released in Anymail v7.1

PR #185, opened 2020-06-03
fix Django3.0 deprecation warning: use force_str instead of force_text on Py3

si14:
Django3.0 comes with the following deprecation warning:

```
/usr/local/lib/python3.7/site-packages/anymail/utils.py:150: RemovedInDjango40Warning: force_text() is deprecated in favor of force_str().
```

This PR fixes the warning by using `force_str` when run under Python 3.


medmunds:
Thanks!

PR #193, opened 2020-07-25
Drop Python 2 and Django 1.11 support

medmunds:
(This is work in progress, and will be released as a breaking change in Anymail v8.)


medmunds:
Released in Anymail 8.0

PR #204, opened 2020-09-11
Call SparkPost HTTP API directly (drop unmaintained library)

### LINKED ISSUE #203: Call SparkPost API directly (drop unmaintained library)
Up to v7.x, Anymail has integrated with SparkPost through their "super-mega-official" [sparkpost Python package](https://pypi.org/project/sparkpost/).

Although it still seems to be working for Anymail, that package hasn't been updated in over 2.5 years, has accumulated a number of bugs and merged-but-unreleased PRs, and is apparently "[no longer supported](https://github.com/SparkPost/python-sparkpost/issues/196#issuecomment-593980026)" by the company.

SparkPost has a straightforward, JSON based [sending API](https://developers.sparkpost.com/api/transmissions/). It would be simple to call it directly (through requests) like Anymail does with nearly every other ESP. Going through the official Python library doesn't really add any value, and actually complicates ongoing maintenance. (Anymail contains its own workarounds for some of the library issues.)

An added benefit of calling the API directly is that newer SparkPost features (e.g., subaccounts, A/B testing) would become immediately available through Anymail's `esp_extra`, without having to wait for a library update (that may never be coming).

This would be a breaking change, as anyone using `esp_extra` with SparkPost would need to update it to the new format. (I'm not sure how common that is with SparkPost.)

medmunds:
Closes #203

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,7 +55,6 @@ jobs:
     # Install without optional extras (don't need to cover entire matrix)
     - { env: TOXENV=django31-py37-none, python: 3.7 }
     - { env: TOXENV=django31-py37-amazon_ses, python: 3.7 }
-    - { env: TOXENV=django31-py37-sparkpost, python: 3.7 }
     # Test some specific older package versions
     - { env: TOXENV=django22-py37-all-old_urllib3, python: 3.7 }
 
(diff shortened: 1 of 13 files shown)


medmunds:
Released in Anymail 8.0

PR #216, opened 2020-12-27
Added missing comma in example code

OmenApps:
This is probably the smallest PR in history, but just adding a missing comma to the docs.


medmunds:
No PR is too small, and every improvement to the docs helps the next person to read them. Thanks!

PR #250, opened 2021-11-11
Update Trove classifiers

Andrew-Chen-Wang:


medmunds:
Thanks for this. 

(I wish there were some way to automate keeping these updated. The problem is PyPI sometimes lags in adding the classifiers for new Django versions, and if you try to publish the package before they've added the classifier it gets rejected.)

Andrew-Chen-Wang:
Yea I don't really know of a way to automate this. Maybe bumpversion? Autocommit GitHub actions with a script?

In terms of when the trove classifiers are added, I believe they are manually added via PRs in the trove repo itself. Last I saw, a Django fellow created a PT for a new classifier when 3.1 alpha came out.

medmunds:
[Released in Anymail v8.5]

PR #276, opened 2022-06-02
Fix: Handle SubscriptionChange event in Postmark webhook

puru02:
Handle SubscriptionChange event in Postmark webhook.
This event was previously not handled and treated as event type Unknown.

The SuppressSending field indicates whether the SubscriptionChange was a deactivation or reactivation.

Postmark doesn't provide a consistent way to distinguish between explicit unsubscribes and bounces so a lot of bounced emails have event type sent to Unknown. This fix will prevent that and classify bounced, unsubscribed, and resubscribed emails correctly.

https://postmarkapp.com/developer/webhooks/subscription-change-webhook

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -25,6 +25,19 @@ Release history
 ^^^^^^^^^^^^^^^
     ..  This extra heading level keeps the ToC from becoming unmanageably long
 
+vNext
+-----
+
+*Unreleased changes*
+
+Fixes
+~~~~~
+
+* **Postmark:** Handle Postmark's SubscriptionChange events as Anymail
+  unsubscribe, subscribe, or bounce tracking events, rather than "unknown".
+  (Thanks to `@puru02`_ for the fix.)
+
+
 v8.6 LTS
 --------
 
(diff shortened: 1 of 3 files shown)


medmunds:
@puru02 thanks for this.

I'd like to have some tests before merging, to cover both deactivation and reactivation events. Also, if you have an example bounce event that's getting miscategorized as unknown right now, it would be good to add a test for that.

If you'd like to add the test cases, you could use the existing test_postmark_webhooks.py [`test_bounce_event`](https://github.com/anymail/django-anymail/blob/v8.6/tests/test_postmark_webhooks.py#L25-L63) as a guide. Or if not, I'm happy to add the tests—just let me know.

medmunds:
@puru02 thanks for this.

I'd like to have some tests before merging, to cover both deactivation and reactivation events. Also, if you have an example bounce event that's getting miscategorized as unknown right now, it would be good to add a test for that.

If you'd like to add the test cases, you could use the existing test_postmark_webhooks.py [`test_bounce_event`](https://github.com/anymail/django-anymail/blob/v8.6/tests/test_postmark_webhooks.py#L25-L63) as a guide. Or if not, I'm happy to add the tests—just let me know.

puru02:
@medmunds I've added tests for bounce, unsubscribe, and resubscribe events

medmunds:
@puru02 looks great. Thanks!

PR #277, opened 2022-06-03
Support requests session customization, document use for automatic retries

dgilmanAIDENTIFIED:
The requests library lets you mount adapters to configure policy such as timeouts and retries. Here's an attempt at an API that extends anymail to let you do so. This is the initial draft, I am looking forward to feedback on how this should be done. I also skipped writing docs.

The functionality is exposed through a new ANYMAIL_REQUESTS_ADAPTERS configuration value in the Django settings:

```
ANYMAIL_REQUESTS_ADAPTERS = [
    ["https://foobar.com", HTTPAdapter(pool_connections=100)],
]
```

### DIFF
--- a/anymail/backends/base_requests.py
+++ b/anymail/backends/base_requests.py
@@ -25,17 +25,12 @@ class AnymailRequestsBackend(AnymailBaseBackend):
             return False  # already exists
 
         try:
-            self.session = requests.Session()
-        except requests.RequestException:
+            self.session = self.create_session()
+        except Exception:
             if not self.fail_silently:
                 raise
-        else:
-            self.session.headers["User-Agent"] = "django-anymail/{version}-{esp} {orig}".format(
-                esp=self.esp_name.lower(), version=__version__,
-                orig=self.session.headers.get("User-Agent", ""))
-            if self.debug_api_requests:
-                self.session.hooks['response'].append(self._dump_api_request)
-            return True
+
+        return True
 
     def close(self):
         if self.session is None:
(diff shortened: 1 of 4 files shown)


medmunds:
@dgilmanAIDENTIFIED cool, thanks for this! You've integrated it nicely into Anymail.

My one concern is whether there are a bunch of other requests.Session options we'd want to expose in a similar way. E.g., `cert` for a custom client certificate, or `proxies`. (And if so, what happens in future requests releases?) I'd like to avoid Anymail having a lot of requests-specific settings.

A more general approach could be a "session creation hook", which Anymail would call with the new session. Clients could do anything they want to that session, including mounting transport adapters. I'd probably lean toward Django's signal receiver mechanism to implement the hook (Anymail already uses it for some [other hook-type features](https://anymail.dev/en/stable/sending/signals/)), but a setting could work, too.

Or one other way to modify the requests.Session—and this works in the current Anymail—is subclassing. You can subclass any Anymail requests-based backend to modify the session after it's created (or even create the session in a completely different way), and just set Django's EMAIL_BACKEND to your subclass. Right now, you have to override `AnymailRequestsBackend.open()`, which is a little ugly. It would probably be good to extract an `AnymailRequestsBackend.create_session()` method with base implementation `return requests.Session()`, just to facilitate this sort of override.

dgilmanAIDENTIFIED:
Those are all good ideas. If you can flip a coin or whatever and pick one I'll give it a shot. I think any of these are good ways forward as long as they're documented.

dgilmanAIDENTIFIED:
My big use case is automatic retries on network errors. I was planning on documenting it on the one doc page that mentions how to catch network errors and retries in a background task.

anymail/backends/base_requests.py
```diff
@@ -20,21 +20,26 @@ def __init__(self, api_url, **kwargs):
         super().__init__(**kwargs)
         self.session = None
 
-    def open(self):
-        if self.session:
-            return False  # already exists
-
+    def create_session(self):
```
medmunds, line 23: Please move `create_session()` down with the other overridable methods (after `_send()`), and add a doc comment that it must return a requests.Session and indicating subclasses can override to augment the session, but should generally call the super method to create it.

medmunds (changes requested):
Nice! I didn't know about retryable adapters.

Couple of minor changes requested in review comments.

medmunds (approved):

medmunds:
Looks great. Thank you for the contribution!

PR #284, opened 2022-08-08
Confirm support for Django 4.1.

tim-schilling:
Handles the deprecation of django.utils.timezone.utc
~~This required aliasing utc dynamically based on the version of
Django in the utils module.~~

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -36,6 +36,8 @@ Features
 * Support customizing the requests.Session for requests-based backends,
   and document how this can be used to mount an adapter that simplifies
   automatic retry logic. (Thanks to `@dgilmanAIDENTIFIED`_.)
+* Confirm support for Django 4.1 and resolve deprecation warning regarding
+  ``django.utils.timezone.utc``.
 
 Fixes
 ~~~~~
--- a/anymail/backends/base.py
+++ b/anymail/backends/base.py
@@ -1,5 +1,5 @@
 import json
-from datetime import date, datetime
+from datetime import date, datetime, timezone
 
 from django.conf import settings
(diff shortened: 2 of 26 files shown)


medmunds:
I think we could just switch to using Python's datetime.timezone.utc unconditionally. (The Django version is a holdover from Python 2 compatibility.)

tim-schilling:
@medmunds that sounds reasonable, but I think dropping versions of python and django should be done in a separate PR. I wrote this one up to maintain all compatibilities with python versions and django versions that the project had listed.

medmunds:
@tim-schilling Anymail dropped support for Python 2.7 back in v8.0 (2020-09-11), and the setup.py requirements and trove classifiers are all Python 3.5 or later since that time. (I'm not sure where you saw earlier versions listed in the project, if something was overlooked please do open a bug report.)

I'd like to avoid version-conditional code paths unless there's just no other solution. Since datetime.timezone.utc has been available since Python 3.2, it would be cleaner to just use that everywhere, rather than introducing a new .utilts.utc that switches based on Django versions.

adamchainz:
+1, I don't see any reason not to use `datetime.timezone.utc` unconditionally.

tim-schilling:
Working on getting those changed now.

medmunds:
Thanks for this!

tim-schilling:
@medmunds as a gentle nudge, when you have time can you issue a new version with this change?

medmunds:
@tim-schilling sorry for the delay on this (and on your PR - thanks for those updates). Been away for a couple of months and am slowly getting caught up. I'm planning to get to a couple of minor issues and then release within the next week.

PR #292, opened 2022-12-10
Fix link to mailjet apikeys

JelteF:
The old link returns a 404 now


medmunds:
@JelteF thanks for this update

Issue #327, opened 2023-07-20
[Inbound] Apple Mail inline attachments not taken into account


martinezleoml:
Hi there!

We've figured out that inbound email processing of emails sent through Apple Mail app (tested on macOS) resulted on inline attachments (like images) not being taken into account at all by django-anymail.

- Anymail version: 10.0
- ESP: AWS SES
- Django version: 4.1.3
- macOS version: Ventura 13.4.1
- Apple Mail version: 16.0

### Reproduction

1. Send a basic email with picture(s) using Apple Mail:
![CleanShot 2023-07-20 at 17 26 22@2x](https://github.com/anymail/django-anymail/assets/759215/3a54f2f4-7648-48af-b86c-5b9929dbe0e3)

2. Copy-paste the resulting raw message content (sent by Apple Mail) into a file named `message.raw`:
```text
Return-Path: <redacted@example.com>
Received: from smtpclient.apple ([REDACTED])
        by smtp.gmail.com with ESMTPSA id h9-REDACTED.88.2023.07.20.08.16.22
        for <redacted@example.com>
        (version=TLS1_2 cipher=ECDHE-ECDSA-AES128-GCM-SHA256 bits=128/128);
        Thu, 20 Jul 2023 08:16:22 -0700 (PDT)
From: "REDACTED" <redacted@example.com>
Content-Type: multipart/mixed; boundary="Apple-Mail=_B8C68BCC-061A-452F-A5CD-474B952DC8B5"
Mime-Version: 1.0 (Mac OS X Mail 16.0 \(3731.600.7\))
Subject: Apple Mail attachment
Message-Id: <FEFBB6BE-E215-403D-8FC8-38A51F0F07F3@example.com>
Date: Thu, 20 Jul 2023 17:16:12 +0200
To: "REDACTED" <redacted@example.com>
X-Mailer: Apple Mail (2.3731.600.7)

--Apple-Mail=_B8C68BCC-061A-452F-A5CD-474B952DC8B5
Content-Transfer-Encoding: 7bit
Content-Type: text/plain; charset=us-ascii

Some text before rectangle


--Apple-Mail=_B8C68BCC-061A-452F-A5CD-474B952DC8B5
Content-Disposition: inline; filename=rectangle.png
Content-Type: image/png; x-unix-mode=0644; name="rectangle.png"
Content-Transfer-Encoding: base64

[...]

PR #328, opened 2023-07-20
fix: handle attachments from Apple Mail client

martinezleoml:
This fixes #327 (email attachments from Apple Mail client not detected).

PR #330, opened 2023-07-25
Inbound: better handle inline parts

martinezleoml:
Relates to #327.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,21 +33,32 @@ vNext
 Features
 ~~~~~~~~
 
-* **Inbound:** Introduce `inlines` property to better handle message's inline parts, especially
-  inline attachments that are not featured with a `Content-ID` header. This better complies with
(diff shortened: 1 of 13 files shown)


medmunds:
Wow, this looks great! Thanks.

My only suggestion would be to add a test for the new `inlines` list property, verifying that it includes inline parts both with and without content-ids (to cover the Apple Mail case you originally reported).

martinezleoml:
Good catch!
I've added a test in a separate commit (feel free to squash it or event edit my previous commits if needed).

By the way, thanks a lot for being reactive on reviewing this. ;)

medmunds:
Thanks.

In the docs, I updated a bunch of `email.message.Message` → `EmailMessage` refs and added some Sphinx "versionadded" and "versionchanged" info; I also fixed some now-outdated comments in the code; and credited you in the changelog.

medmunds:
Oops, closes #327. (And #328, I guess.)

martinezleoml:
Thanks for the credits and for your help. ;)

Issue #349, opened 2024-01-09
Add support to Unisender Go


Arondit:
https://go.unisender.ru/

It has transactional sending API and supports event tracking webhooks. It has a free tier, but only if you can prove your domain. 

I already have all needed code and preparing pull request.

medmunds:
Merged in #352.

Thanks again for the implementation

medmunds:
(Released in Anymail 10.3)

PR #352, opened 2024-01-10
Unisender Go: new ESP

Arondit:
https://github.com/anymail/django-anymail/issues/349

- [x] Backend
- [x] Custom headers for metadata and tags
- [x] Tracking webhook
- [x] Integration tests
- [x] Docs
- [x] Readme/project/tox/workflows


medmunds:
Thanks for this!

I'll take a closer look over the weekend

Arondit:
Sorry, this time it must pass.

medmunds:
Sorry, I don't know why it requires me to re-approve the tests workflow every time.

You can run the tests on your own machine with `python runtests.py` in the project root. More options: https://anymail.dev/en/stable/contributing/#testing

Incidentally, I'm trying to get an anymail.dev subdomain set up in a Unisender Go account to run integration tests, but they seem to have some problems validating subdomains. (And in general, I'm seeing a lot of 50x 
and 429 errors just trying to navigate their dashboard—though I suppose that might be some sort of geopolitical filtering issue, since I'm connecting from the U.S.)

anymail/webhooks/unisender_go.py
```diff
@@ -0,0 +1,172 @@
...
+            event_type=event_type,
+            timestamp=timestamp_utc,
+            message_id=metadata.get("message_id", None),
+            event_id=None,
+            recipient=event_data["email"],
+            reject_reason=anymail_reject_reason,
+            mta_response=delivery_info.get("destination_response", ""),
+            tags=None,
+            metadata=metadata,
+            click_url=None,
```
medmunds, line 169: Looks like `event_data.get("url")` would provide `click_url` for the "clicked" event.

anymail/webhooks/unisender_go.py
```diff
@@ -0,0 +1,172 @@
...
+            user_agent=delivery_info.get("use_ragent", ""),
```
medmunds, line 170: Typo in `user_agent`

tests/test_unisender_go_payload.py
```diff
@@ -0,0 +1,138 @@
...
+SUBSTITUTION_ONE = {"arg1": "arg1"}
+SUBSTITUTION_TWO = {"arg2": "arg2"}
+
+
+class TestUnisenderGoPayload(SimpleTestCase):
+    @override_settings(
+        ANYMAIL_UNISENDERGO_SKIP_UNSUBSCRIBE=False,
+        ANYMAIL_UNISENDERGO_API_KEY=None,
+        ANYMAIL_UNISENDERGO_API_URL="",
+    )
```
medmunds, line 26: (You can apply `@override_settings` at the class level, if every test in the class wants these overrides.)

docs/esps/index.rst
```diff
@@ -35,33 +36,33 @@ The table below summarizes the Anymail features supported for each ESP.
 
 .. rst-class:: sticky-left
 
-============================================  ============  =======  ============  ===========  ==========  ===========  ==========  ==========  ========  ==========  ===========
-Email Service Provider                        |Amazon SES|  |Brevo|  |MailerSend|  |Mailgun|    |Mailjet|   |Mandrill|   |Postal|    |Postmark|  |Resend|  |SendGrid|  |SparkPost|
-============================================  ============  =======  ============  ===========  ==========  ===========  ==========  ==========  ========  ==========  ===========
+============================================  ============  =======  ============  ===========  ==========  ===========  ==========  ==========  ========  ==========  ==============  ===========
+Email Service Provider                        |Amazon SES|  |Brevo|  |MailerSend|  |Mailgun|    |Mailjet|   |Mandrill|   |Postal|    |Postmark|  |Resend|  |SendGrid|  |Unisender Go|  |SparkPost|
```
medmunds, line 40: Please keep ESP names in alphabetic order (so "Unisender Go" would come last in the current list, after "SparkPost").

(Also, sorry this table is so hard to edit.)

anymail/webhooks/unisender_go.py
```diff
@@ -0,0 +1,172 @@
...
+}
+"""
+
+
+class UnisenderGoTrackingWebhookView(AnymailCoreWebhookView):
+    """Handler for UniSender delivery and engagement tracking webhooks"""
+
+    esp_name = "UnisenderGo"
+    signal = tracking
+
```
medmunds, line 74: Probably want to set `warn_if_no_basic_auth = False`, since MD5 signature validation is sufficient.

(Example of [webhook that uses signature verification](https://github.com/anymail/django-anymail/blob/main/anymail/webhooks/mailgun.py#L36-L62).)

medmunds (changes requested):
Thanks again for this PR!

I've added several comments/requests. Most of them are about making things more consistent with other Anymail backends.

Arondit:
Hello again. Thank you for comments, it will take a bit of time to fix them all. They also require adding new tests for extra parameters, which i did not include before. I will fix them all and update PR on Monday.

anymail/backends/unisender_go.py
```diff
@@ -0,0 +1,286 @@
...
+        self.data["global_metadata"] = metadata
+
+    def set_send_at(self, send_at: datetime.datetime) -> None:
+        self.data.setdefault("options", {})["send_at"] = send_at
+
+    def set_tags(self, tags: dict[str, str]) -> None:
+        self.data["tags"] = tags
+
+    def set_template_id(self, template_id: str) -> None:
+        self.data["template_id"] = template_id
```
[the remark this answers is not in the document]
Arondit: 1. I moved extra fields to "set_esp_extra"
2. Added some fields there
3. I did NOT move template_id, because i saw set_template_id in some other ESPs, so i decided, it is OK. 

I dont check this conversation as resolved 'cause not sure, if you approve it.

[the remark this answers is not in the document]
Arondit: Oh, now I understand, thank you. Yeah, it's very nice and flexible, cool. I've been using the library configs kinda wrong way for a while xD

Arondit:
I don't really understand now, what went wrong again. Locally everything works fine. Also the error is weird, `AttributeError: module 'importlib' has no attribute 'util'`. But python's had importlib.util since 3.3. I have no idea, how i could break it.  
Sorry for so much mess-up.

medmunds:
@Arondit the `'importlib' has no attribute 'util'` error isn't a problem in your code, it's a [problem in the `hatch` packaging tool](https://github.com/pypa/hatch/issues/1221) that Anymail uses. One of hatch's dependencies changed out from under it a few days ago, in a way that surfaced this bug in hatch. (We run weekly builds to catch problems like this; tomorrow's test run would have noticed it, but your PR got there first.)

There's a PR just opened in hatch to fix the problem, but if that doesn't make it into a release soon, I'll try to update Anymail to work around it.

Until then, if the tests run fine on your local machine, that should be good.

medmunds:
Tests are working again. (Bug in hatch was fixed and a patch released last night.)

Arondit:
Hello again. So, tests are working, all checks are passed and comments are closed. Is there anything else I should do? Also, did it work with integration tests for this ESP?

medmunds:
There are still a handful of unresolved change requests from earlier. (They might be hiding behind a not-so-helpful GitHub "load more" link like this
image
in the discussion above.)

The two most important ones are handling for `cc` and `bcc` recipients (which I think will currently overwrite the `to` list), and parsing the send status out of Unisender Go's API response.

Also, while the unit tests are helpful, it might be a good idea to also implement at least one or two basic functional tests using the `RequestsBackendMockAPITestCase`, to verify that the API key, url, and basic payload structure are as expected—and more importantly, that future changes don't break that. (Here's an [example from the Postmark tests](https://github.com/anymail/django-anymail/blob/main/tests/test_postmark_backend.py#L59-L75). And [another that would have caught the `cc` issue](https://github.com/anymail/django-anymail/blob/main/tests/test_postmark_backend.py#L77-L95).)

medmunds:
@Arondit I'd be happy to work on resolving the remaining handful of issues from the original review, unless you are already working on them. Just let me know.

Also, do you know if it's true that Unisender Go does not support "cc" or "bcc"? (Or did I miss something in their docs?)

Arondit:
@medmunds i'm working on it, just was busy at work those days, sorry for 2 weeks of silence. 

About "bcc" and "cc". Yes, you're right. They don't support it through their WEB API. There is a possibility to use SMTP API instead, which supports "cc" and "bcc" in a strict mode. Is it a problem?

medmunds:
@Arondit no worries, and no hurry, just didn't want to conflict with any work you might be doing.

On cc/bcc, I was asking related to [this review comment](https://github.com/anymail/django-anymail/pull/352#discussion_r1454204741). We can just handle cc/bcc as unsupported features. (Rather than trying to use their SMTP API. Anymail's current design doesn't really work with SMTP.)

medmunds:
[Updated to pick up changes from main branch]

Arondit:
Fix all missed comments (yeah, you're right, GitHub've hidden them from me).

medmunds:
Thanks, it's looking great. I have a few things I wanted to double check, and then can get this merged.

I added some live integration tests. If you want to try them locally:

```sh
export ANYMAIL_TEST_UNISENDER_GO_API_KEY="your-api-key"
export ANYMAIL_TEST_UNISENDER_GO_API_URL="https://go1.unisender.ru/ru/transactional/api/v1"  # or "go2"
export ANYMAIL_TEST_UNISENDER_GO_DOMAIN="your.validated.sending.domain"
# optional:
export ANYMAIL_TEST_UNISENDER_GO_TEMPLATE_ID="id of email template in your account"

python runtests.py tests.test_unisender_go_integration
```

(The integration tests send actual email to an anymail.dev null mailbox, so each run will use up about five sends from your quota.)

medmunds (approved):

Arondit:
After small fixes tests worked. Thank you. Also, '/' in the end of ANYMAIL_TEST_UNISENDER_GO_API_URL 
 appeared to be significant. 

Снимок экрана 2024-02-22 в 23 55 34

medmunds:
Fixed a handful of issues and consistency with other ESP backends.

I just want to update the docs a little and then will merge.

medmunds:
Hmm, it looks like cc and bcc *are* possible through the web API: https://godocs.unisender.ru/cc-and-bcc. (I think they just updated the API docs—I don't remember seeing that a few weeks ago.)

Basically, we'd need to generate an entry in `"recipients"` for *each* of the *to*, *cc*, and *bcc* emails, and then construct `"To"` and `"CC"` headers that list *all* of their respective addresses. Unisender Go (effectively) just treats those headers as text to include verbatim in the generated message, without trying to parse them.

There might be some tricky cases to consider when using Anymail's batch sending mode. (I *think* for batch send we just skip the common "To" header, but keep the common "CC" header. But need to think about this a little more.) 

Also we'd want to test whether non-ASCII display names require special handling.

Arondit:
Add one unit test and fix integration test (bcc address needs to be sent on approved domain). Also thank you very much for work around added cc/bcc feature. I started to do the same today earlier and noticed, you'd already done it, but better. I guess, now it is ready to merge, isn't it?

medmunds:
Good catch on the integration test, thanks.

I emailed Unisender Go tech support earlier today about the display-name quoting bugs and a couple other issues. Hoping to hear back from them once they get into the office, but if not I'll go ahead and merge this with the workarounds.

medmunds:
@Arondit OK, this is merged and will be in the next release!

Thanks again for all your work on this.

Arondit:
nice, thank you

PR #354, opened 2024-02-01
MailPace: New ESP

paul-oms:
Hi Anymail team, thanks for building such well designed and easy to extend software! We're an ESP (https://mailpace.com) that would like to be fully out of the box with you.

Let us know if there's more we need to do, or improvements to the code that we can make.

- [x] Backend
- [x] Webhooks
- [x] Inbound
- [x] Tags
- [x] Tests
- [x] Docs
- [x] Readme/project/tox/workflows etc.


paul-oms:
More issues than expected on the other Python versions - we'll fix them locally and re-open a PR once complete (with signed commits)

medmunds:
Thanks, looking forward to it!

One problem causing multiple failing tests is the optional library: anymail.webhooks.mailpace needs to be importable whether or not pynacl is installed. There's a decent example of how to handle that in [anymail.webhooks.resend](https://github.com/anymail/django-anymail/blob/main/anymail/webhooks/resend.py#L14-L23).

paul-oms:
Thanks! I've attempted that pattern in the new PR 🤞

PR #355, opened 2024-02-01
MailPace: New ESP

paul-oms:
Hi Anymail team, a fresh PR for https://mailpace.com with verified commits, pynacl optional, and the linting issues fixed. I was unable to run everything locally, so 🤞 this passes the Github Actions this time round.

- [x] Backend
- [x] Webhooks
- [x] Inbound
- [x] Tags
- [x] Tests
- [x] Docs
- [x] Readme/project/tox/workflows etc.


tests/utils_mailpace.py
```diff
@@ -0,0 +1,56 @@
...
+    def set_private_key(self, private_key):
+        self.private_key = private_key
+
+    def post(self, *args, **kwargs):
+        data = kwargs.get("data", "").encode("utf-8")
+
+        headers = kwargs.setdefault("headers", {})
+        if "X-MailPace-Signature" not in headers:
+            signature = sign(self.private_key, data)
+            headers["X-MailPace-Signature"] = signature
```
medmunds, line 49: The `headers` param was added in Django 4.2. 

You might borrow [this code](https://github.com/anymail/django-anymail/blob/v10.2/tests/test_resend_webhooks.py#L59-L64) (add to the super call) to backport headers for earlier Django versions.

tests/utils_mailpace.py
```diff
@@ -0,0 +1,56 @@
...
+
+        webhook_key = derive_public_webhook_key(self.private_key)
+        with override_settings(ANYMAIL={"MAILPACE_WEBHOOK_KEY": webhook_key}):
```
medmunds, line 52: `override_settings` won't merge dicts, so this will blow away any other `ANYMAIL` settings you might have from the test cases. An alternative is to use the non-dict setting name: `override_settings(ANYMAIL_MAILPACE_WEBHOOK_KEY=webhook_key)`.

medmunds:
This is looking pretty good so far. Added a couple of comments above that might help get the tests working. I'll try to get to a closer review over the weekend.

paul-oms:
Phew, finally got the tests to pass. Let me know what else is needed.

Also, if you need an API key for integration testing, set up an account at https://app.mailpace.com and we'll add you to the free plan (presumably I can't put the API key anywhere for you).

medmunds:
Nice! Glad you got the tests working.

I created mailpace.anymail.dev as a testing domain. It would be great if you could set it up for free API use.

There's a decision to be made about tracking webhook verification. I think the choices are:
A. Signature verification (MAILPACE_WEBHOOK_KEY, pynacl) is optional; if not used, HTTP basic auth (WEBHOOK_SECRET) is "required" (strongly encouraged via Anymail warning message). This is how you've got it documented.
B. Signature verification is required. It's an error to use the MailPace tracking webhook without MAILPACE_WEBHOOK_KEY set *and* pynacl installed. This is easier to implement and test.

For option A, we need to pull in something like [this code](https://github.com/anymail/django-anymail/blob/v10.2/anymail/webhooks/resend.py#L48-L55) to control the basic auth warning, and then also rework the webhook tests to cover both scenarios (rather than just skipping all tests when pynacl isn't installed), similar to how the [Resend webhook tests](https://github.com/anymail/django-anymail/blob/v10.2/tests/test_resend_webhooks.py) handle optional signature validation with the svix package.

Option B would also be fine—it's really up to you. (pynacl is not tied to any proprietary service, seems likely to be maintained for many years, and doesn't have any other dependencies. It wasn't immediately obvious svix fit those criteria, which is why Anymail's Resend webhooks make signature verification with svix optional.) For option B, we just need to add [this line](https://github.com/anymail/django-anymail/blob/v10.2/anymail/webhooks/mailersend.py#L26), and allow the configuration error if the webhook key is missing.

paul-oms:
Thanks! I think Option A is the right choice - always better to make 3rd party dependencies optional.

Does this change cover it?

paul-oms:
> I created mailpace.anymail.dev as a testing domain. It would be great if you could set it up for free API use.

Added to our secret free plan ;)

anymail/backends/mailpace.py
```diff
@@ -0,0 +1,190 @@
...
+        """Init options from Django settings"""
+        esp_name = self.esp_name
+        self.server_token = get_anymail_setting(
+            "server_token", esp_name=esp_name, kwargs=kwargs, allow_bare=True
+        )
+        api_url = get_anymail_setting(
+            "api_url",
+            esp_name=esp_name,
+            kwargs=kwargs,
+            default="https://app.mailpace.com/api/v1/send",
```
medmunds, line 24: For consistency with other ESPs, I'd suggest just using the API base url here (`"https://app.mailpace.com/api/v1/"`), and then implementing MailPacePayload.get_api_endpoint that just returns `"send"`. ([Example](https://github.com/anymail/django-anymail/blob/v10.2/anymail/backends/resend.py#L83-L84).) Even though "send" is your only API right now, this simplifies things if you later decide to add a separate API like "send-template" or "send-batch".

anymail/backends/mailpace.py
```diff
@@ -0,0 +1,190 @@
...
+        return MailPacePayload(message, defaults, self)
+
+    def raise_for_status(self, response, payload, message):
+        # We need to handle 400 responses in parse_recipient_status
+        if response.status_code != 400:
+            super().raise_for_status(response, payload, message)
+
+    def parse_recipient_status(self, response, payload, message):
+        # Prepare the dict by setting everything to queued without a message id
+        unknown_status = AnymailRecipientStatus(message_id=None, status="queued")
```
medmunds, line 40: Shouldn't this be `status="unknown"`? (And then it gets overwritten below for recipients whose status is known.)

anymail/backends/mailpace.py
```diff
@@ -0,0 +1,190 @@
...
+
+        if status_msg == "queued":
+            # Add the message_id to all of the recipients
+            for recip in payload.to_cc_and_bcc_emails:
+                recipient_status[recip.addr_spec] = AnymailRecipientStatus(
+                    message_id=id, status="queued"
+                )
+        elif status_msg == "error":
+            if "errors" in json_response:
+                for field in ["to", "cc", "bcc"]:
```
medmunds, line 75: I'm a little confused by the logic in this section.

The goal is to provide a specific send status for each recipient, if known. 

I think this code will end up setting the sent status for *every* recipient based on the last reported error in the response (for *any* recipient). So if all "to" addresses were valid, but one of the cc's was on the blocked list, all of the "to" and the "cc" recipients would get reported as "blocked".

(Are there any cases where MailPace would send the message to some recipients, but not to others? Or if there's a problem with *any* to/cc/bcc recipient, is the entire message always rejected?)

anymail/backends/mailpace.py
```diff
@@ -0,0 +1,190 @@
...
+class MailPacePayload(RequestsPayload):
+    def __init__(self, message, defaults, backend, *args, **kwargs):
+        headers = {
+            "Content-Type": "application/json",
+            "Accept": "application/json",
+        }
+        self.server_token = backend.server_token  # esp_extra can override
+        self.to_cc_and_bcc_emails = []
+        self.merge_data = None
+        self.merge_metadata = None
```
medmunds, line 126: (There's no need to have `merge_data` and `merge_metadata` properties if MailPace doesn't support batch merge.)

anymail/backends/mailpace.py
```diff
@@ -0,0 +1,190 @@
...
+            self.to_cc_and_bcc_emails += emails
+
+    def set_subject(self, subject):
+        self.data["subject"] = subject
+
+    def set_reply_to(self, emails):
+        if emails:
+            reply_to = ", ".join([email.address for email in emails])
+            self.data["replyto"] = reply_to
+
```
medmunds, line 161: I'd suggest implementing `set_extra_headers` to allow *List-Unsubscribe*. Something like:

```python
    def set_extra_headers(self, headers):
        if "list-unsubscribe" in headers:
            self.data["list_unsubscribe"] = headers.pop("list-unsubscribe")
        if headers:
            self.unsupported_features("extra_headers (other than List-Unsubscribe)")
```

(The `headers` param is a CaseInsensitiveDict.)

anymail/backends/mailpace.py
```diff
@@ -0,0 +1,190 @@
...
+            self.data["attachments"] = [
+                self.make_attachment(attachment) for attachment in attachments
+            ]
+
+    def set_tags(self, tags):
+        if tags:
+            if len(tags) == 1:
+                self.data["tags"] = tags[0]
+            else:
+                self.data["tags"] = tags
```
medmunds, line 190: Please also implement `set_esp_extra`. This helps future-proof Anymail when MailPace adds new options to the send API.

(Also, you have code above that supports overriding the server token on individual messages, via esp_extra, so that needs to be handled here. I'm guessing this is important for domain-specific API keys. You can copy from the [Postmark backend](https://github.com/anymail/django-anymail/blob/v10.2/anymail/backends/postmark.py#L386-L389), which has exactly the same feature.)

anymail/webhooks/mailpace.py
```diff
@@ -0,0 +1,145 @@
...
+
+    webhook_key = None
+
+    def __init__(self, **kwargs):
+        try:
+            self.webhook_key = get_anymail_setting(
+                "webhook_key", esp_name=self.esp_name, kwargs=kwargs, allow_bare=True
+            )
+            self.warn_if_no_basic_auth = False
+        except AnymailConfigurationError:
```
medmunds, line 59: You can pass `default=None` to `get_anymail_setting` to avoid the exception, and then just check if the returned value is None.

anymail/webhooks/mailpace.py
```diff
@@ -0,0 +1,145 @@
...
+    # When/if MailPace does this, move this to the parent class
+    def validate_request(self, request):
+        if self.webhook_key:
+            try:
+                signature_base64 = request.headers["X-MailPace-Signature"]
+                signature = base64.b64decode(signature_base64)
+            except (KeyError, binascii.Error):
+                raise AnymailWebhookValidationFailure(
+                    "MailPace webhook called with invalid or missing signature"
+                )
```
medmunds, line 87: Suggest `raise ... from error` here for cleaner stack trace. (Anymail isn't totally consistent about this yet, but I'm trying to use this approach in new and updated code.)

anymail/webhooks/mailpace.py
```diff
@@ -0,0 +1,145 @@
...
+            message = request.body
+
+            try:
+                verify_key.verify(message, signature)
+            except (CryptoError, ValueError):
+                raise AnymailWebhookValidationFailure(
+                    "MailPace webhook called with incorrect signature"
+                )
+        else:
+            return True
```
medmunds, line 102: There's no need to return anything. `validate_request` either raises an exception or it doesn't; the return value is ignored.

anymail/webhooks/mailpace.py
```diff
@@ -0,0 +1,145 @@
...
+    signal = inbound
+
+    def esp_to_anymail_event(self, esp_event):
+        # Use Raw MIME based on guidance here:
+        # https://github.com/anymail/django-anymail/blob/main/ADDING_ESPS.md
+        message = AnymailInboundMessage.parse_raw_mime(esp_event.get("raw", None))
+
+        return AnymailInboundEvent(
+            event_type=EventType.INBOUND,
+            timestamp=timezone.now(),
```
medmunds, line 141: Please use `timestamp=None` if the inbound payload doesn't provide an event timestamp. (Timestamp needs to be unchanged if the same event is received twice.)

medmunds (changes requested):
This looks really good!

I have a handful of feature requests and a handful of minor suggestions in this review.

The updated webhook validation looks fine. I'll take a closer look at the tests tomorrow, but I think you've got all the cases covered.

anymail/backends/mailpace.py
paul-oms: It's the latter - if there's a problem with any recipient (e.g. blocked or invalid) the email will be rejected (error http status from the endpoint) and not sent, as the entire email is technically not valid in that case.

paul-oms:
All done - there's one point to check above, but I think it's correct as is.

medmunds:
Was working on integration tests, and noticed there seems to be a problem with commas in "to" display names. Try sending `"to": "\"Example, Inc.\" <info@example.com>"`. I get a 400 response with `"errors": { "to": [ "is invalid" ] }`.

(No need to go through Anymail—same problem just using curl with json. Probably also affects cc, bcc, and reply_to, though I haven't tested. The "from" field seems OK with commas in display names.)

Trouble with various characters in email display names is pretty common, and normally Anymail implements a [workaround involving RFC-2047 encoding](https://github.com/anymail/django-anymail/blob/v10.2/anymail/backends/resend.py#L96-L117). Which is an option here, but maybe you want to look into a fix on your end instead?

medmunds:
Also, here are some integration tests: [test_mailpace_integration.py](https://github.com/anymail/django-anymail/files/14201884/test_mailpace_integration.py.txt)

You can run them locally with:

```shell
export ANYMAIL_TEST_MAILPACE_SERVER_TOKEN=your-server-token
export ANYMAIL_TEST_MAILPACE_DOMAIN=sending.domain.for.your.server.token
python runtests.py tests.test_mailpace_integration 
```

(These tests won't run in this PR until it's merged, because repository secrets aren't available within PRs from forks. They send real messages, but the destination addresses are all sink emails at anymail.dev.)

(Unfortunately, I can't just add this file to your PR because of a [long-standing GitHub bug with forks scoped to organizations](https://github.com/orgs/community/discussions/5634). You might want to [add me as a collaborator](https://github.com/orgs/community/discussions/5634#discussioncomment-2155236) on your django-anymail fork.)

paul-oms:
> Was working on integration tests, and noticed there seems to be a problem with commas in "to" display names. Try sending `"to": "\"Example, Inc.\" <info@example.com>"`. I get a 400 response with `"errors": { "to": [ "is invalid" ] }`.
> 
> (No need to go through Anymail—same problem just using curl with json. Probably also affects cc, bcc, and reply_to, though I haven't tested. The "from" field seems OK with commas in display names.)
> 
> Trouble with various characters in email display names is pretty common, and normally Anymail implements a [workaround involving RFC-2047 encoding](https://github.com/anymail/django-anymail/blob/v10.2/anymail/backends/resend.py#L96-L117). Which is an option here, but maybe you want to look into a fix on your end instead?

Thanks! This is most likely a bug on our end. I'll look into it and see if we can fix it.

medmunds:
Thanks, pushed the integration tests.

You might want to add some tests for the features you enabled recently: list-unsubscribe header (but other headers unsupported) and [overriding the api token in esp_extra](https://github.com/anymail/django-anymail/blob/v10.2/tests/test_postmark_backend.py#L731-L750). (I'm happy to add these tests, too.)

I'm still thinking about the best way to map recipient status, will follow up later. Everything else looks good.

anymail/backends/mailpace.py
medmunds: OK, thanks, that helps. I also did a little testing. [Sorry, this comment is kind of long—there are a bunch of cases to consider.]

The goal of `parse_recipient_status` is: based on the API response, for *each* individual "to" recipient, provide the send status *if we know it*, or "failed" if the message definitely wasn't sent to (queued for) that recipient but we can't exactly say why, or "unknown" if we're not sure whether or not the message was sent to that particular recipient. (This is especially important with some other ESPs that don't fail the entire message, but instead send to some recipients but not others.)

If that status information is also available for "cc" and "bcc" recipients, it's great to provide it. Otherwise, it's OK to just omit cc and bcc recipients.

In my testing, it looks like MailPace's API response currently reports:
* `{"errors":{"to":["is invalid"]}}` if at least one (but not necessarily all) of the "to" recipients is invalid
* `{"errors":{"to":["contains a blocked address"]}}` if at least one (but not necessarily all) of the "to" recipients is on the blocked list
* `{"errors":{"to":["contains a blocked address", "is invalid"]}}` if both conditions apply
* With multiple "to" addresses, there is no way to tell from the API response which of the addresses did or didn't cause these errors
* It seems like errors in "cc" or "bcc" addresses are not currently reported—the message is queued and delivered to the "to" recipients. [Aside: it looks like delivery is attempted for a blocked cc. An invalid cc field seems to cause the message to get stuck in the Queued status. I'm guessing these might be bugs?]

With that in mind, here are some scenarios `parse_recipient_status` needs to cover:

* One or more "to" recipients, no errors
  * Current implementation: all "to" recipients are reported "queued", which is the desired result
* Exactly one "to" recipient, invalid address
  * Current implementation: the recipient is reported as "invalid", which is the desired result
* Exactly one "to" recipient, on the blocked list 
  * Current implementation: the recipient is reported as "rejected", which is the desired result
* Multiple "to" recipients, some invalid and/or blocked (and some not)
  * Current implementation: *all* "to" recipients are reported as either "invalid" or "rejected", even if that individual recipient was valid. This is misleading. (Also, when both errors apply, whichever happens to be last in the errors list determines the status for all recipients.)
  * Desired result: with multiple recipients and some invalid/blocked errors, use status "failed" for all "to" recipients. (We know the message wasn't sent to that recipient, but for each individual address we can't be sure why—or even that there was a problem with *that* address.)
* Any of the above, but also with "cc" and "bcc" recipients
  * Current implementation: "cc" and "bcc" recipients are given the same status as the "to" recipient(s) ("queued" or "invalid" or "rejected"), which may be misleading (because MailPace doesn't currently report errors on cc and bcc, and the cc/bcc recipients aren't necessarily invalid/blocked just because one of the to recipients is)
  * Desired result: probably easiest to just omit "cc" and "bcc" recipients. (Could instead give them status "unknown" if the "to" is "queued", or "failed" if the "to" is invalid/blocked, but that doesn't add much value.)
* Zero "to" recipients, one or more "cc" and/or "bcc" recipients
  * [Developers sometimes try to use cc/bcc-only messages to implement something like a mailing list]
  * Current implementation: all "cc" and "bcc" recipients are given status "invalid", which will be misleading (the cc and bcc recipients are likely valid email addresses)
  * Desired result: since MailPace doesn't support messages without "to", just let its `{"errors":{"to":["undefined field"]}}` response be reported as an AnymailRequestsAPIError. (Don't try to handle this through recipient status.)
* More than 50 "to" recipients
  * Current implementation: all recipients are reported as "invalid", which is misleading
  * Desired result: just let MailPace's `"number of email addresses exceeds maximum volume"` response be reported as an AnymailRequestsAPIError. (Don't try to handle this through recipient status.)

Here's a suggested approach that I think achieves all of this:
* Replace `to_cc_and_bcc_emails` with just `to_emails`
* If `response.status_code == 200`, return `status = "queued"` for all `to_emails`
  * Is it necessary to check `parsed_response["status"] == "queued"`? (Could there ever be any other status with a 200 response?)
* If `response.status_code == 400` and the *only* field in `parsed_response["errors"]` is `"to"` and the errors in `parsed_response["errors"]["to"]` are *only* `"is invalid"` and/or `"contains a blocked address"`:
  * If there is exactly one `to_emails`:
    * If `"is invalid"`, return `status = "invalid"` for all `to_emails`
    * If `"contains a blocked address"`, return `status = "rejected"` for all `to_emails`
    * If both, return `status = "failed"` for all `to_emails`
  * Else (more than one `to_emails`) return `status = "failed"` for all `to_emails` (we don't know which are invalid or rejected)
* In *all other* cases, raise an `AnymailRequestsAPIError` for the response
  * This includes all other problems that might be reported in `parsed_response["errors"]["to"]`
  * This also covers the case where MailPace later starts reporting errors in cc/bcc fields
* (Wrap all `parsed_response[...]` access to convert KeyError and TypeError to `AnymailRequestsAPIError("Unexpected MailPace API response format"...)`)

Does that make sense? Have I missed anything?

anymail/webhooks/mailpace.py
```diff
@@ -0,0 +1,140 @@
...
+            if event_type == EventType.REJECTED
+            else RejectReason.BOUNCED
+            if event_type == EventType.BOUNCED
+            else None
+        )
+        tags = payload.get("tags", [])
+
+        return AnymailTrackingEvent(
+            event_type=event_type,
+            timestamp=parse_datetime(payload["created_at"]),
```
medmunds, line 115: `timestamp` is meant to be the time the event occurred. I think created_at is the time the email was originally sent, even for later events like delivered or bounced. I wonder if updated_at might be more appropriate?

anymail/webhooks/mailpace.py
```diff
@@ -0,0 +1,140 @@
...
+            event_id=payload["id"],
```
medmunds, line 116: `event_id` is meant to be a unique ID for this event, to help avoid processing the same event twice. I don't believe MailPace provides that sort of ID, so probably just pass `event_id=None`

anymail/webhooks/mailpace.py
```diff
@@ -0,0 +1,140 @@
...
+            message_id=payload["message_id"],
```
medmunds, line 117: `message_id` is meant to match the `message_id` returned from the backend's `parse_recipient_status`, so that users can match up webhook events with the original sent message. So I think we want MailPace's "internal email id" here: `message_id=payload["id"]`.

anymail/webhooks/mailpace.py
```diff
@@ -0,0 +1,140 @@
...
+            recipient=payload["to"],
```
medmunds, line 118: `recipient` must be only the addr_spec portion of the (one) relevant recipient address.  It looks like `payload["to"]` is the entire *To* header, including display names and possibly multiple recipients.

You should be able to use something like `parse_address_list(payload["to"])[0].addr_spec` to get the first addr_spec in the list. (That function is from anymail.utils. If that expression raises a ValueError or IndexError, the to field is unparseable or empty, and it's fine to pass `recipient=None`.)

anymail/webhooks/mailpace.py
```diff
@@ -0,0 +1,140 @@
...
+            tags=tags,
+            reject_reason=reject_reason,
```
medmunds, line 120: Please also add `esp_event=esp_event` to the `AnymailTrackingEvent` constructor. This allows handler functions access to the complete raw event if necessary.

medmunds (changes requested):
Handful of issues from testing event tracking webhooks

anymail/backends/mailpace.py
paul-oms: Thanks @medmunds  - my suggestion here is for us to fix the issues with non-reporting of errors and blocked emails to cc and bcc'd recipients first (plus the other bug reported below), then come back and fix the PR appropriately. We'll get back shortly on that.

medmunds: > fix the issues with non-reporting of errors and blocked emails to cc and bcc'd recipients first

Sounds good, thanks.

medmunds:
[Rebased to pick up main branch changes]

PR #357, opened 2024-02-19
Brevo: add batch send support

### LINKED ISSUE #353: Support Brevo batch sending with merge_data
Hi,

I'm reaching out to report a potential issue I encountered while reading the Anymail docs (for use with Brevo). This is my first time working with both Anymail and Brevo, so I might be getting it all wrong.

The documentation states ([here](https://anymail.dev/en/stable/esps/brevo/#batch-sending-merge-and-esp-templates)) that Brevo does not support per-recipient variables, and we shouldn't use `merge_data`.

From what I understood from the Brevo documentation ([here](https://developers.brevo.com/docs/batch-send-transactional-emails#creating-the-api-call-with-a-template-id)), it looks like per-recipient params are supported (using `params` in each subobjects of `messageVersions`).

As of my writing, this section of the Brevo documentation was updated one month ago. Is this a feature that was recently supported, or am I misunderstanding something?

medmunds:
Closes #353

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -25,6 +25,18 @@ Release history
 ^^^^^^^^^^^^^^^
     ..  This extra heading level keeps the ToC from becoming unmanageably long
 
+vNext
+-----
+
+*unreleased changes*
+
+Features
+~~~~~~~~
+
+* **Brevo:** Add support for batch sending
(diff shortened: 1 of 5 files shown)


medmunds:
(Released in Anymail 10.3)

PR #359, opened 2024-02-20
Resend: support batch send

### LINKED ISSUE #356: Add batch sending support for Resend
Resend [recently announced](https://resend.com/blog/introducing-the-batch-emails-api) a new [batch sending API](https://resend.com/docs/api-reference/emails/send-batch-emails). Anymail should add support for it.

Resend's batch send API basically takes an array the same payload structure used for the regular send API. So it:
- Wouldn't support Anymail's `merge_data` (because Resend doesn't support ESP-rendered templates even for regular send)
- *Would* support Anymail's `merge_metadata` (batch send supports custom headers, which Anymail uses for metadata and tags with Resend)
- Doesn't support attachments "yet" (need to test if attachments field causes error, or is just silently ignored right now)

medmunds:
Add support for `merge_metadata` and new Resend email/batch API.

Closes #356

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -35,6 +35,8 @@ Features
 
 * **Brevo:** Add support for batch sending
   (`docs <https://anymail.dev/en/latest/esps/brevo/#batch-sending-merge-and-esp-templates>`__).
+* **Resend:** Add support for batch sending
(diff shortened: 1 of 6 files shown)


medmunds:
(Released in Anymail 10.3)

PR #364, opened 2024-03-12
Brevo: Rename SendinBlue to Brevo

medmunds:
I've been rethinking my arguments in #321 about leaving "SendinBlue" in the code while calling it "Brevo" everywhere else. I now feel this creates more confusion (particularly with new users) than it avoids.

This PR renames "sendinblue" to "brevo" throughout the code, but maintains compatibility versions on the old names and URLs (with deprecation warnings). 

The compatibility versions use the old `esp_name`, which means that settings names and other related features are also unchanged. So if you have `EMAIL_BACKEND = "…⁠sendinblue.EmailBackend"`, Anymail will continue to look for `SENDINBLUE_API_KEY` in the settings. But when you update to `EMAIL_BACKEND = "…⁠brevo.EmailBackend"`, you'll need to switch to `BREVO_API_KEY` at the same time.

I don't have any particular schedule for removing the deprecated compatibility versions. (They don't really add maintenance overhead.)

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -40,6 +40,7 @@ jobs:
         # combination, to avoid rapidly consuming the testing accounts' entire send allotments.
         config:
           - { tox: django41-py310-amazon_ses, python: "3.10" }
+          - { tox: django41-py310-brevo, python: "3.10" }
           - { tox: django41-py310-mailersend, python: "3.10" }
           - { tox: django41-py310-mailgun, python: "3.10" }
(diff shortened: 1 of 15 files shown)


medmunds:
[Split into two commits, which makes it possible to see the actual diffs in the renamed files before adding the compatibility versions on the original filenames: backends/sendinblue.py → backends/brevo.py; webhooks/sendinblue.py → webhooks/brevo.py]

medmunds:
(Released in Anymail 10.3)

PR #370, opened 2024-04-18
Fix reference to non-existent method on `EmailMessage`

dmwyatt:
It's `get_all`, not `getall`.  https://docs.python.org/3.12/library/email.message.html#email.message.EmailMessage.get_all

I checked python versions 3.6-3.12 and confirmed it's always been the case.


medmunds:
Thanks, good catch.

PR #371, opened 2024-05-16
Support merge_headers in Amazon SES bulk send

carrerasrodrigo:
Please check  ReplacementHeaders
https://docs.aws.amazon.com/ses/latest/APIReference-V2/API_SendBulkEmail.html

This headers are very useful for the new requirements of email clients like gmail, yahoo, etc. 

Allows to add headers as a template. This enables, for example, to add an unique unique unsubscribe header for each email.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -41,6 +41,13 @@ Breaking changes
   (Anymail 10.0 switched to the SES v2 API by default. If your ``EMAIL_BACKEND``
   setting has ``amazon_sesv2``, change that to just ``amazon_ses``.)
 
+Features
+~~~~~~~~
+
+* **Amazon SES:** Add new ``merge_headers`` option for per-recipient
+  headers with template sends. (Requires boto3 >= 1.34.98.)
+  (Thanks to `@carrerasrodrigo`_ the implementation.)
+
 
 v10.3
 -----
@@ -1615,6 +1622,7 @@ Features
 .. _@Arondit: https://github.com/Arondit
 .. _@b0d0nne11: https://github.com/b0d0nne11
 .. _@calvin: https://github.com/calvin
+.. _@carrerasrodrigo: https://github.com/carrerasrodrigo
 .. _@chrisgrande: https://github.com/chrisgrande
 .. _@cjsoftuk: https://github.com/cjsoftuk
 .. _@costela: https://github.com/costela
--- a/anymail/backends/amazon_ses.py
+++ b/anymail/backends/amazon_ses.py
@@ -298,6 +298,9 @@ class AmazonSESV2SendEmailPayload(AmazonSESBasePayload):
         # metadata.
         self.mime_message["X-Metadata"] = self.serialize_json(metadata)
 
+    def set_merge_headers(self, merge_headers):
+        self.unsupported_feature("merge_headers without template_id")
+
     def set_tags(self, tags):
         # See note about Amazon SES Message Tags and custom headers in set_metadata
         # above. To support reliable retrieval in webhooks, use custom headers for tags.
@@ -327,9 +330,6 @@ class AmazonSESV2SendEmailPayload(AmazonSESBasePayload):
     def set_merge_data(self, merge_data):
         self.unsupported_feature("merge_data without template_id")
 
-    def set_merge_header_data(self, merge_data):
-        self.unsupported_feature("merge_header_data without template_id")
-
     def set_merge_global_data(self, merge_global_data):
         self.unsupported_feature("global_merge_data without template_id")
 
@@ -342,7 +342,7 @@ class AmazonSESV2SendBulkEmailPayload(AmazonSESBasePayload):
         # late-bind recipients and merge_data in finalize_payload
         self.recipients = {"to": [], "cc": [], "bcc": []}
         self.merge_data = {}
-        self.merge_header_data = {}
+        self.merge_headers = {}
 
     def finalize_payload(self):
         # Build BulkEmailEntries from recipients and merge_data.
@@ -359,8 +359,9 @@ class AmazonSESV2SendBulkEmailPayload(AmazonSESBasePayload):
             ]
 
         # Construct an entry with merge data for each "to" recipient:
-        self.params["BulkEmailEntries"] = [
-            {
+        self.params["BulkEmailEntries"] = []
+        for to in self.recipients["to"]:
+            entry = {
                 "Destination": dict(ToAddresses=[to.address], **cc_and_bcc_addresses),
                 "ReplacementEmailContent": {
                     "ReplacementTemplate": {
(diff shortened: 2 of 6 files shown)


carrerasrodrigo:
Please let me know if any change is needed in order to merge it! Thanks a lot for your time!

medmunds:
Thanks for this! Yes, very relevant with the need for List-Unsubscribe headers.

Two requests:
1. I'd prefer naming the new attribute `merge_headers`, to parallel Anymail's existing `merge_metadata` (which is the per-recipient version of `metadata`). 
2. If there are no merge headers for a particular recipient, don't add the ReplacementHeaders field to the SES API call at all (rather than adding it with an empty list). This makes it easier to release without requiring a boto3 upgrade. (If people aren't using merge_headers, their current boto3 version will continue to work without error.)

Also, it would be good to test merge_headers in the Amazon SES integration tests, but this will need to wait for a boto3 release that supports them.

I can look into supporting merge_headers for other ESPs, too. It seems like Brevo, Mailjet, Postmark, Resend, and Sendgrid all have reasonably-easy support for per-recipient headers in their batch APIs. Unisender Go supports them (using substitutions), but only for the List-Unsubscribe header. Mailgun may or may not support them. (I can't tell from Mailgun's docs if their "replacement variables" work in header values; will need to test.)

carrerasrodrigo:
Thanks a lot for your feedback @medmunds! I made the changes as requested. Let me know any additional modification!
I think I will have time to checkout Sendgrid and add an additional PR 🤞

medmunds:
Excellent, thanks.

Sorry, one more thing: just noticed this is expecting the `merge_headers` for each recipient in the `[{"Name": header, "Value": value}, ...]` list form used by SES. I think it would be more natural to use a `{header: value, ...}` dict, matching [Django's EmailMessage `headers`](https://docs.djangoproject.com/en/5.0/topics/email/#emailmessage-objects:~:text=headers%3A%20A%20dictionary%20of%20extra%20headers%20to%20put%20on%20the%20message):

```python
    message = AnymailMessage(
        ...,
        headers={
            "List-Unsubscribe-Post": "List-Unsubscribe=One-Click",
        },
        merge_headers={
            "alice@example.com": {
                "List-Unsubscribe": "<https://example.com/a/>",
            },
            "nobody@example.com": {
                "List-Unsubscribe": "<mailto:unsubscribe@example.com>", 
            },
        },
    )
```

[Btw, we try to use example.com or a similar [reserved example or test domain](https://en.wikipedia.org/wiki/Example.com). xyz.com is an actual domain that belongs to someone.]

carrerasrodrigo:
@medmunds pushed the changes as requested. 

I did not use headers fields, I got this error:
```
======================================================================
ERROR: test_template (tests.test_amazon_ses_backend.AmazonSESBackendAnymailFeatureTests)
With template_id, Anymail switches to SESv2 SendBulkEmail
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/rodrigocarreras/WORK/django-anymail/.tox/django42-py38-all/lib/python3.8/site-packages/django/test/utils.py", line 461, in inner
    return func(*args, **kwargs)
  File "/Users/rodrigocarreras/WORK/django-anymail/tests/test_amazon_ses_backend.py", line 680, in test_template
    message.send()
  File "/Users/rodrigocarreras/WORK/django-anymail/.tox/django42-py38-all/lib/python3.8/site-packages/django/core/mail/message.py", line 299, in send
    return self.get_connection(fail_silently).send_messages([self])
  File "/Users/rodrigocarreras/WORK/django-anymail/anymail/backends/base.py", line 117, in send_messages
    sent = self._send(message)
  File "/Users/rodrigocarreras/WORK/django-anymail/anymail/backends/amazon_ses.py", line 78, in _send
    return super()._send(message)
  File "/Users/rodrigocarreras/WORK/django-anymail/anymail/backends/base.py", line 146, in _send
    payload = self.build_message_payload(message, self.send_defaults)
  File "/Users/rodrigocarreras/WORK/django-anymail/anymail/backends/amazon_ses.py", line 96, in build_message_payload
    return AmazonSESV2SendBulkEmailPayload(message, defaults, self)
  File "/Users/rodrigocarreras/WORK/django-anymail/anymail/backends/base.py", line 340, in __init__
    setter(value)
  File "/Users/rodrigocarreras/WORK/django-anymail/anymail/backends/base.py", line 412, in process_extra_headers
    self.set_extra_headers(headers)
  File "/Users/rodrigocarreras/WORK/django-anymail/anymail/backends/amazon_ses.py", line 447, in set_extra_headers
    self.unsupported_feature("extra_headers with template")
  File "/Users/rodrigocarreras/WORK/django-anymail/anymail/backends/base.py", line 360, in unsupported_feature
    raise AnymailUnsupportedFeature(
anymail.exceptions.AnymailUnsupportedFeature: Amazon SES does not support extra_headers with template
```

code

```
headers={
            "List-Unsubscribe-Post": "List-Unsubscribe=One-Click",
        },
```

Let me know if i'm missing something.

medmunds:
I think my suggestion may have been unclear. I believe the new `merge_headers` Anymail option should take a dict of dicts, not a dict of lists. The desired format is:

```python
        merge_headers={
            "alice@example.com": {  
                # not a list, just a dict of field: value
                "List-Unsubscribe": "<https://example.com/a/>",
                "X-Some-Other-Header": "some-other-value",
            },
            "nobody@example.com": {
                "List-Unsubscribe": "<mailto:unsubscribe@example.com>", 
            },
        },
```

This gives it the same structure as Django's own email `headers` (a.k.a. `extra_headers`), and Anymail's `merge_data` and `merge_metadata`.

Then in the [SES finalize_payload implementation](https://github.com/anymail/django-anymail/blob/25ec0c464a6e76dc6eae4a8484a3c2ce804771d6/anymail/backends/amazon_ses.py#L376-L377), we would convert that dict to the list format ses::SendBulkEmail requires. Something like:

```python
            if to.addr_spec in self.merge_headers:
                entry["ReplacementHeaders"] = [
                    {"Name": key, "Value": value} 
                    for key, value in self.merge_headers[to.addr_spec].items()
                ]
```

Does that seem reasonable?

As for this error…

> `AnymailUnsupportedFeature: Amazon SES does not support extra_headers with template`

…that's [this code](https://github.com/anymail/django-anymail/blob/25ec0c464a6e76dc6eae4a8484a3c2ce804771d6/anymail/backends/amazon_ses.py#L446-L447). Because when it was first implemented, there wasn't a way to pass *any* email headers to ses::SendBulkEmail. We could support it now, using ReplacementHeaders. 

I can add that at some point later, or if you're interested, the idea is that the ReplacementHeaders for each recipient would be the merger of any extra_headers (for all recipients) plus any merge_headers (for that recipient).

carrerasrodrigo:
@medmunds sorry for the misunderstanding, I updated the code to a dict of dicts.  

> I can add that at some point later, or if you're interested, the idea is that the ReplacementHeaders for each recipient would be the merger of any extra_headers (for all recipients) plus any merge_headers (for that recipient). 

I would try to add a new PR supporting this feature 💪 

Hope is ready to merge! have a great day!

medmunds:
Excellent! Thanks again for this—very nice addition.

I'll open a couple of issues to track related work.

PR #376, opened 2024-05-22
Fix broken sendgrid documentation links

GDay:
Thanks for this package!

Some of the SendGrid links are now not available anymore (404 error). Looks like they migrated the docs over to Twilio now. This PR fixes them.


medmunds:
Thanks for this—fighting linkrot in the docs is a constant battle.

(I guess Twilio hasn't heard that [cool URIs don't change](https://www.w3.org/Provider/Style/URI.html).)

PR #379, opened 2024-06-14
SendGrid: better handling of event="dropped", reason="Bounced Address" webhook events.

vitaliyf:
According https://www.twilio.com/docs/sendgrid/for-developers/tracking-events/event#dropped the correct "reason" value for bouncing addresses is "Bounced Address" and that seems to be what I am seeing in SendGrid logs. Without this change those events were being normalized by anymail to `type=dropped, reject_reason=other` and not `reject_reason=bounced`.

I used sample event from their docs, and kept original mapping value since I'm not sure if that may actually show up in other situations.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -52,6 +52,13 @@ Features
   and ``tags`` when sending with a ``template_id``.
   (Requires boto3 v1.34.98 or later.)
 
+Fixes
+~~~~~
+
+* **SendGrid:** In the tracking webhook, correctly report "bounced address"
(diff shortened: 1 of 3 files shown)


medmunds:
Thanks for this. 

The tests are all based on actual webhook payloads observed when SendGrid support was first implemented (because the specific payload values weren't documented). It seems like they've added some values since then, but I agree it doesn't hurt to keep supporting the old "bounced" reason as well.

vitaliyf:
Thank *you* for this project - I was only making tweaks in an inherited codebase, and it was great to discover that we used this nice library that encapsulates such secondary functionality as webhook handling.

Issue #385, opened 2024-07-15
Brevo Webhook Mapping Updates


originell:
Dear Team, first of all: great package. Been using it forever without a fail. Forever grateful!

While debugging an issue with the Brevo (ex SendInBlue) Webhooks, I've discovered 2 slightly related things I'd like to get your opinions about.

#### A) New Events PR 

I noticed that Brevo now has a couple of new events that are not yet mapped in the [webhook](https://github.com/anymail/django-anymail/blob/136ba59ceb5c48048cfbde5a9bf60f323fb798b7/anymail/webhooks/brevo.py#L43).

From what I've [seen their docs](https://developers.brevo.com/docs/transactional-webhooks), the new ones are:

- `complaint`
- `error`
- `proxy_open`

I would suggest the following mapping updates:

- `complaint`: `(EventType.COMPLAINED, RejectReason.SPAM)`
- `error`: `(EventType.FAILED, None)`
- `proxy_open`: `(EventType.OPENED, None)`

If these seem OK to you as well, I'd be happy to supply a PR as well. Just wanted to get a quick OK before investing the time :)

#### B) The "request" Event Question

Looking at the [webhook view's event type mapping](https://github.com/anymail/django-anymail/blob/136ba59ceb5c48048cfbde5a9bf60f323fb798b7/anymail/webhooks/brevo.py#L43), I see this mapping here:

```python
    event_types = {
# ...
        "request": (EventType.QUEUED, None),
# ...
    }
```

I was wondering why this is mapped to QUEUED and not SENT? If you can still remember.

The current docs seem to suggest that the `request` payload is meant as `sent`.

---

Thank you for taking the time maintaining this great package and also keeping the code as neat as it is!

medmunds:
Thanks for catching these new Brevo events. Yes, a PR would be welcome. Your proposed mapping looks correct to me.

[...]

Issue #344, opened 2023-12-07
Inaccuracy in documentation about Brevo's behaviour


thibault:
Hi,

We are using Anymail's status tracking webhook to count the number of times sent emails are opened.

However, an issue with first openings not being counted led me to find something that seems to be inaccurate in anymail's documentation.

Here is an excerpt from [Brevo's related documentation](https://anymail.dev/en/latest/esps/brevo/):

> If you are interested in tracking opens, note that Brevo has both a “First opening” and an “Opened” event type, and will generate both the first time a message is opened. Anymail normalizes both of these events to “opened.” To avoid double counting, you should only enable one of the two.

However, after a few tests, it seems to me that it's just not true.

## Steps to reproduce

- I send an email to my address with Brevo, using any technique (not necessarily anymail)
- I open said email and make sure to load images so the email opening is tracked by Brevo
- I check the logs section in my Brevo's account
- I see a "first opened" event, but no "opened" event at all.

Here, anymail and django versions or any piece of code are irrelevant, because the issue is only with Brevo's behaviour as it is described in the documentation.

Thank you.

medmunds:
@thibault thanks for reporting this. It seems that Brevo has changed how their open webhooks are handled since we first implemented them in Anymail. I am now seeing the same behavior as you. (Also, Brevo seems to have renamed the "Opened" event to "Known open" in their dashboard.)

We should update Anymail's docs to recommend enabling _both_ "Known open" and "First opening" events for open tracking.

(As an aside, I always do a bunch of manual testing with [webhook.site](https://webhook.site) as part of implementing any ESP's webhooks in Anymail, because so far _not a single ESP_ has had complete and accurate documentation for their webhook payloads. I'd like to find a way to add webhooks to Anymail's weekly live integration tests, to catch changes like this one.)

medmunds:
Updated docs in 804cb76aa1935e40ff3cf2976aa05916daac24fc

PR #387, opened 2024-07-21
Brevo webhook updates

originell:
rel #385 

As talked about in the related issue, I've updated the mappings. I also added tests and docs. 

Furthermore, I tried to make the code comment I missed ( "request" event) stand out more, by moving all the comments from inline to above their related entry. I hope this is OK.

Very much looking forward to your comments!

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -26,6 +26,18 @@ Release history
     ..  This extra heading level keeps the ToC from becoming unmanageably long
 
 
+vNext
+-----
+
+*unreleased changes*
+
+Features
+~~~~~~~~
+
+* **Brevo:** Support Brevo's new "Complaint," "Error" and "Loaded by proxy"
+  tracking events. (Thanks to `@originell`_ for the update.)
+
+
 v11.0.1
 -------
 
@@ -1695,6 +1707,7 @@ Features
 .. _@mounirmesselmeni: https://github.com/mounirmesselmeni
 .. _@mwheels: https://github.com/mwheels
 .. _@nuschk: https://github.com/nuschk
+.. _@originell: https://github.com/originell
 .. _@puru02: https://github.com/puru02
 .. _@RignonNoel: https://github.com/RignonNoel
 .. _@sblondon: https://github.com/sblondon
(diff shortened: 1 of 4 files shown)


originell:
A question: while testing the new mappings, I noticed that Brevo had some problems with the proxy_open event. Hence, I contacted their support and I've received this as an answer:

> The team has confirmed that they needed to make a few database changes so that this event type could be added to our collections.
> We have made changes for you and you should now receive “unique_proxy_open” for the first/unique opening.
> For subsequent openings, you will receive “proxy_open” events.

Now, I am asking myself: should I also supply a mapping for `unique_proxy_open` as well? 

It is undocumented and I've only gotten access to this by asking support.

medmunds:
Thanks, your updates look great.

> should I also supply a mapping for `unique_proxy_open` as well?

Yes. (I think.)

I'd guess that `proxy_open` and `unique_proxy_open` will parallel `opened` and `unique_opened`, so mapping both will be necessary to receive all the proxy open events.

We ended up normalizing both of the original "opened" to Anymail's OPENED, and added [this paragraph](https://anymail.dev/en/stable/esps/brevo/#status-tracking-webhooks:~:text=If%20you%20are%20interested,message%20opens%20enable%20both.) to the docs about how to choose. See #344 if you're interested in the history.

> `# Sadly, this is not well documented in the official Brevo API documentation.`

I may get that printed on a T-shirt 😄. (Not to pick on Brevo: pretty much every ESP's documentation for webhooks is some combination of incomplete and just plain inaccurate.)

originell:
Thanks for the feedback! I've now added the suggested changes.

Let me know if something could be done better!

medmunds:
That's great, thanks. I update the docs to combine some of the earlier information about open events with your new information.

(Also, nice work debugging Brevo's proxy open behavior and the missing `unique_proxy_open` event!)

originell:
Perfect! Thanks for merging those. Sounds way better that way.

> nice work debugging Brevo's proxy open behavior
Haha thanks. But that was more people work than debugging work. Their support was very helpful though. Wasn't more than 2-3 emails.

PR #397, opened 2024-08-31
Fix AnymailWebhookValidationFailure on some events

MikeVL:


medmunds:
@MikeVL thank you for the contribution. Could you please open an issue describing the problem this is trying to fix?

(I can try to guess: it seems like Unisender Go uses a fragile webhook authentication methodology that depends on implementation specific details of JSON serialization. But it would be helpful to confirm this and get more details.)

MikeVL:
@medmunds On some events the auth digest check fails. Probably because during deserialization and serialization json keys are sorted in a different order. 

Now this code works fine in my production.

medmunds:
@MikeVL thanks for this.

In trying to update the tests for this PR, I discovered several other problems with the current Unisender Go webhook implementation. I've opened a new PR #401 covering all of them, including a version of your fix from here.

Issue #249, opened 2021-10-06
Inconsistency in testing environment when sending mass emails.


Genetical:
**Python Version:** `3.9.7`
**Django Version:** `3.2.7`
**Requests Version:** `2.26.0`
**Anymail Version:** `8.4`
**ESP:** `Postmark`

When sending an email job to an ESP, you can set the `merge_data` param to be an empty dictionary. This will result in individual emails sent to each recipient in the `to` param. 

If 10 emails are given, each recipient will recieve the email but will only see their own email in the recipient field as opposed to a list of all the recipients. The below code functions as expected when sent to an ESP. 

```python
if merge_data is None:
    merge_data = {}

log.info(f"Email job triggered. Sending to {len(recipients)} emails. Subject line: {subject}")

message = EmailMessage(
    subject=subject, body=body, from_email=sender, to=recipients, reply_to=reply_to
)

message.content_subtype = "html"
message.template_id = template_id
message.merge_data = merge_data
message.merge_global_data = merge_global_data
message.esp_extra = esp_extra

try:
    log.info(f"Sending email request to ESP. Attempt {self.request.retries + 1} of 5.")
    message.send()
except AnymailAPIError as exc:
    log.warning(f"API Error when communicating with ESP: {exc}")
    self.retry(exc=exc, countdown=int(random.uniform(2, 4) ** self.request.retries))
    # Countdown set to an exponential backoff with random jitter.
```

However, when running this code in a test environment (i.e Setting the email backend to `anymail.backends.test.EmailBackend`), only one `EmailMessage` object will be added to `mail.outbox` with all of the recipients in the one email's `to` attr.

Would it be possible for the test email backend to identify when `merge_data` has been defined and to instead generate an individual email for each recipient in the outbox to better replicate how ESP interactions work? Thank you

Genetical:
Thank you for this response. What you said about each message representing an API call makes complete sense, so I understand you not wanting to change it. You could possibly consider creating a separate test backend to better emulate an esp which subclasses the main test backend (in the fashion you demonstrated above) for when people do want to more closely mock an ESP.

Regarding the copy method, I've glanced over the EmailMessage definition and the only alternative I can see is just to construct the messages "manually" in the `post_to_esp` method. Although I'm not sure if that would result in the same value for attributes like the date and id.

[...]

PR #406, opened 2024-11-10
Add support for Mailtrap

cahna:
Hi, and thanks for creating such a great library! I would love to have Mailtrap support, and I figured it would be helpful if I started on that. **This is very much a work-in-progress**, as I'm still trying to understand how to handle all of the features that Anymail provides, and whether those features are supported by Mailtrap. I have already caught at least one category of features that Mailtrap does not support: inbound emails and reply-to.

If anyone would be willing to provide some feedback on what I have, and any guidance on how to continue, I would appreciate it, and I will be sure to add tests and documentation.

I am also having trouble getting local builds and tests to work, or I would have started on adding tests. If someone could point me in the direction of minimal steps to get tests working on Ubuntu, I would appreciate it. (The [contributing docs](https://anymail.dev/en/stable/contributing/#testing) don't seem to work for me).

Cheers


anymail/backends/mailtrap.py
```diff
@@ -0,0 +1,248 @@
...
+            "test_api_url",
+            esp_name=self.esp_name,
+            kwargs=kwargs,
+            default="https://sandbox.api.mailtrap.io/api/",
+        )
+        if not test_api_url.endswith("/"):
+            test_api_url += "/"
+        self.test_api_url = test_api_url
+
+        bulk_api_url = get_anymail_setting(
```
cahna, line 184: I haven't quite determined how Backends should handle bulk emails, but it also does not appear that Mailtrap's APIs are any different for their "Email sending" vs "Bulk email" APIs https://api-docs.mailtrap.io/docs/mailtrap-api-docs/.

Maybe this `self.bulk_api_url` is unnecessary, and it should be up to the user to configure which API url to use via `self.api_url`?

anymail/backends/mailtrap.py
```diff
@@ -0,0 +1,248 @@
...
+            not parsed_response.get("success")
+            or ("errors" in parsed_response and parsed_response["errors"])
+            or ("message_ids" not in parsed_response)
+        ):
+            raise AnymailRequestsAPIError(
+                email_message=message, payload=payload, response=response, backend=self
+            )
+
+        # Not the best status reporting. Mailtrap only says that the order of message-ids will be in this order (but JSON is unordered?)
+        recipient_status_order = [*message.to, *message.cc, *message.bcc]
```
cahna, line 237: Mailtrap's status reporting is pretty poor, and I'm not even sure if this will work consistently, given that it is depending on ORDERED JSON responses :facepalm:

medmunds:
Thanks for this! It looks pretty good at first glance; I'll try to take a closer look later this week.

> Mailtrap does not support: inbound emails and reply-to.

Missing inbound support is not unusual; it's fine to just ignore it.

Reply-to is a little surprising. Some ESPs require handling this as an extra header. (I haven't had a chance to look at Mailtrap's docs.)

> I am also having trouble getting local builds and tests to work

What's going wrong? (In the contributing docs, I notice the "test a representative combination of Python and Django versions" command is outdated—the current version should be `tox -e lint,django51-py312-all,django40-py38-all,docs`. But other than that I'd expect it to work. I'm usually on macOS, but the GitHub tests all run on Ubuntu.)

cahna:
@medmunds 
> Reply-to is a little surprising. Some ESPs require handling this as an extra header.

Aha! I'm no email expert, so TIL! I will add the reply-to-via-headers support momentarily.

cahna:
@medmunds 
> What's going wrong? (In the contributing docs, I notice the "test a representative combination of Python and Django versions" command is outdated—the current version should be tox -e lint,django51-py312-all,django40-py38-all,docs. But other than that I'd expect it to work. I'm usually on macOS, but the GitHub tests all run on Ubuntu.)

Switching to a dev container, using pyenv to install 3.12 and 3.8, and using the updated tox command got further. I got a complaint that python 3.8 was missing, but creating a symlink /usr/bin/python3.8 to the pyenv installed version fixed that. I think I'm good to go for writing tests. Thanks.

anymail/backends/mailtrap.py
medmunds: Yeah, it looks like Mailtrap's ["bulk" vs transactional stream](https://help.mailtrap.io/article/113-sending-streams) is sort of a choice of outbound server pool. (Postmark and Amazon SES have similar concepts. They allow choosing the stream at send time by, respectively, providing a different API key or a different "configuration set" payload option.)

> Maybe this `self.bulk_api_url` is unnecessary, and it should be up to the user to configure which API url to use via `self.api_url`?

I think that's correct. Make transactional the default api_url, and users who want some sends to go through the bulk API can override it. (Globally via settings, or for specific sends using [`get_connection()`][1]. Plus there's an upcoming Django "EMAIL_PROVIDERS" feature that will make this even easier.)

[1]: https://anymail.dev/en/stable/installation/#anymail-settings-reference:~:text=you%20can%20override%20most%20settings%20on%20a%20per%2Dinstance%20basis%20by%20providing%20keyword%20args%20where%20the%20instance%20is%20initialized%20(e.g.%2C%20in%20a%20get_connection()%20call%20to%20create%20an%20email%20backend%20instance

medmunds (changes requested):
@cahna this is looking really solid! I have a bunch of suggestions, some responses to your questions, and some questions of my own.

I haven't been through the webhook you just added yet; will take a look at that soon.

[GitHub sometimes [hides review comments](https://github.com/orgs/community/discussions/130618): please search the page for "Hidden Conversations" to be sure you see them all.]

tests/test_mailtrap_webhooks.py
```diff
@@ -0,0 +1,375 @@
...
+
+
+@tag("mailtrap")
+class MailtrapWebhookSecurityTestCase(WebhookBasicAuthTestCase):
+    def call_webhook(self):
+        return self.client.post(
+            "/anymail/mailtrap/tracking/",
+            content_type="application/json",
+            data=json.dumps({}),
+        )
```
medmunds, line 20: When content_type is application/json, Django's test Client.post will handle json serialization for you:

```suggestion
        return self.client.post(
            "/anymail/mailtrap/tracking/",
            content_type="application/json",
            data={},
        )
```

(This was added in Django 2.1, but some of Anymail's older tests haven't been updated to remove the unnecessary json.dumps().)

tests/test_mailtrap_webhooks.py
```diff
@@ -0,0 +1,375 @@
...
+                    "email": "receiver@example.com",
+                    "event_id": "bede7236-2284-43d6-a953-1fdcafd0fdbc",
+                },
+            ]
+        }
+        response = self.client.post(
+            "/anymail/mailtrap/tracking/",
+            content_type="application/json",
+            data=json.dumps(payload),
+        )
```
medmunds, line 46:
```suggestion
        response = self.client.post(
            "/anymail/mailtrap/tracking/",
            content_type="application/json",
            data=payload,
        )
```

(See earlier note. Same also applies to other post() calls below.)

medmunds (changes requested):
OK, the webhooks look really good, too. Handful of minor comments.

Thanks again for your work on this.

anymail/backends/mailtrap.py
cahna: Will do. Do you think there would be any value in sorting the recipients by email address before sending?

cahna:
Thanks for your feedback @medmunds! Taking a look now.

FYI: I only have Sundays available to work on this for the foreseeable future.

pyproject.toml
```diff
@@ -63,6 +64,7 @@ dependencies = [
     "django>=4.0",
     "requests>=2.4.3",
     "urllib3>=1.25.0",  # requests dependency: fixes RFC 7578 header encoding
+    "typing_extensions>=4.12",  # for older Python compatibility
```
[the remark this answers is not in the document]
cahna: I'm happy to implement whatever you request. I have started to rely heavily on Python type checking for my own development, but if you want me to remove all type hints once the work part is done, I'm happy to do that, too.

I confirm your assessment that it has only been added for `TypedDict` and `NotRequired` support in versions <3.11.

> at the very least we could probably make this a version-dependent dependency: "typing_extensions>=4.12; python_version < 3.11"

I will do this for now.

anymail/backends/mailtrap.py
```diff
@@ -0,0 +1,257 @@
...
+
+    def set_template_id(self, template_id):
+        # Mailtrap requires integer (not string) TemplateID:
+        self.data["template_id"] = template_id
+
+    def set_merge_data(self, merge_data):
+        self.data.setdefault("template_variables", {}).update(merge_data)
+
+    def set_merge_global_data(self, merge_global_data):
+        self.data.setdefault("template_variables", {}).update(merge_global_data)
```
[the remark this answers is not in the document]
cahna: > Mailtrap doesn't seem to support per-recipient merge data

After re-reviewing their docs, I think you are correct. That's unfortunate. I have removed the `set_merge_data` method, and added a `Dict[str, Any]` type hint to `merge_global_data`.

anymail/backends/mailtrap.py
```diff
@@ -0,0 +1,257 @@
...
+        "to": NotRequired[List[MailtrapAddress]],
+        "cc": NotRequired[List[MailtrapAddress]],
+        "bcc": NotRequired[List[MailtrapAddress]],
+        "attachments": NotRequired[List[MailtrapAttachment]],
+        "headers": NotRequired[Dict[str, str]],
+        "custom_variables": NotRequired[Dict[str, str]],
+        "subject": str,
+        "text": str,
+        "html": NotRequired[str],
+        "category": NotRequired[str],
```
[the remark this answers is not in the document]
cahna: I did this, but apparently forgot to commit it :sweat_smile:. Fixed, and thanks for catching.

anymail/webhooks/mailtrap.py
```diff
@@ -0,0 +1,101 @@
...
+        "unsubscribe": RejectReason.UNSUBSCRIBED,
+        "reject": RejectReason.BLOCKED,
+        "suspension": RejectReason.OTHER,
+        "soft bounce": RejectReason.OTHER,
+    }
+
+    def esp_to_anymail_event(self, esp_event: MailtrapEvent):
+        event_type = self.event_types.get(esp_event["event"], EventType.UNKNOWN)
+        timestamp = datetime.fromtimestamp(esp_event["timestamp"], tz=timezone.utc)
+        reject_reason = self.reject_reasons.get(esp_event["event"], RejectReason.OTHER)
```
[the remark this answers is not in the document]
cahna: Thanks for catching this.

pyproject.toml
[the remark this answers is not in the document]
cahna: I'm not sure what the best practice is for type hints in libs that support multiple python versions, but me intuition tells me there has to be a better way than this(?):

```python3
if sys.version_info < (3, 11):
    from typing_extensions import Any, Dict, List, Literal, NotRequired, TypedDict
else:
    from typing import Any, Dict, List, Literal, NotRequired, TypedDict
```

I suppose if #394 is coming, then maybe it is worth making `typing_extensions` a requirement? That way all of the type imports can come from `typing_extensions` instead of `typing`:

```python3
from typing_extensions import Any, Dict, List, Literal, NotRequired, TypedDict
```

I will do some more research on this.

cahna:
I will start working on integration tests locally with my own free account. @medmunds, would you be able to setup a free mailtrap account, put the API key and test inbox ID into the repo's github secrets, and expose them to the build as environment variables?

FYI: the free mailtrap tier allows 100 test emails/month.

anymail/backends/mailtrap.py
cahna: Removed `bulk_api_url`

anymail/backends/mailtrap.py
```diff
@@ -0,0 +1,272 @@
...
+    def build_message_payload(self, message, defaults):
+        return MailtrapPayload(message, defaults, self)
+
+    def parse_recipient_status(
+        self, response, payload: MailtrapPayload, message: AnymailMessage
+    ):
+        parsed_response = self.deserialize_json_response(response, payload, message)
+
+        # TODO: how to handle fail_silently?
+        if not self.fail_silently and (
```
cahna, line 247: While trying to duplicate test coverage from the other backend tests, I found that the fail_silently test(s) don't work, and this doesn't seem to be the correct way to do this. How should I handle that? Should I?

cahna:
@medmunds I got started on backend unit tests, but I have left several `@unittest.skip("TODO: ...")` on tests that either might not be relevant for Mailtrap, or that have outstanding questions on how to adjust the backend code to make the test pass.

anymail/backends/mailtrap.py
medmunds: I would not try to sort recipients. I don't think it would help anything, and there might be cases where senders are deliberately putting addresses in a specific order.

medmunds:
> @medmunds, would you be able to setup a free mailtrap account, put the API key and test inbox ID into the repo's github secrets, and expose them to the build as environment variables?

Done: variables `ANYMAIL_TEST_MAILTRAP_DOMAIN` and `ANYMAIL_TEST_MAILTRAP_TEST_INBOX_ID` and secret `ANYMAIL_TEST_MAILTRAP_API_TOKEN` are installed. (Secrets aren't available to PR builds, so the integration tests won't actually run until it's merged into main.)

anymail/backends/mailtrap.py
```diff
@@ -0,0 +1,272 @@
...
+            ]
+            recipient_status = {
+                email: AnymailRecipientStatus(
+                    message_id=message_id,
+                    status="sent",
+                )
+                for email, message_id in zip(
+                    recipient_status_order, parsed_response["message_ids"]
+                )
+            }
```
medmunds, line 270: Here's an approach for reporting invalid responses. It's also probably worth checking the number of message_ids, since zip() won't raise an error for that. 

```suggestion
        try:
            message_ids = parsed_response["message_ids"]
        except KeyError as err:
            raise AnymailRequestsAPIError(
                "Invalid response format: missing message_ids",
                email_message=message, payload=payload, response=response, backend=self
            ) from err

        # message-ids will be in this order
        recipient_status_order = [
            *payload.recipients_to,
            *payload.recipients_cc,
            *payload.recipients_bcc,
        ]
        if len(recipient_status_order) != len(message_ids):
            raise AnymailRequestsAPIError(
                "Invalid response format: wrong number of message_ids",
                email_message=message, payload=payload, response=response, backend=self
            )
            
        recipient_status = {
            email: AnymailRecipientStatus(
                message_id=message_id,
                status="sent",
            )
            for email, message_id in zip(recipient_status_order, message_ids)
        }
```

API error responses should already be covered by raise_for_status() in the superclass. (But if you wanted to add a check like `if parsed_response.get("errors") or not parsed_response.get("success"): raise AnymailRequestsAPIError("Unexpected errors in {response.status_code} response", ...)`, that wouldn't hurt anything.)

tests/test_mailtrap_backend.py
```diff
@@ -0,0 +1,299 @@
+# FILE: tests/test_mailtrap_backend.py
```
medmunds, line 1: ??? (is this some editor convention?)

tests/test_mailtrap_backend.py
```diff
@@ -0,0 +1,299 @@
...
+
+    def test_send_with_invalid_api_token(self):
+        """Test sending an email with an invalid API token"""
+        self.set_mock_response(status_code=401, raw=b'{"error": "Invalid API token"}')
+        with self.assertRaises(AnymailAPIError):
+            self.message.send()
+
+    @unittest.skip("TODO: is this test correct/necessary?")
+    def test_send_with_recipients_refused(self):
+        """Test sending an email with all recipients refused"""
```
medmunds, line 112: This is probably unnecessary, since Mailtrap doesn't seem to provide per-recipient send status. 

Mailtrap's error documentation isn't all that descriptive, so we might need to test this. If you try to send a message with two recipients, both on your suppression list (or previous bounces), what happens? (Send API success and later webhook reject? Or send API error?)

Similarly, if you send with one valid recipient and one suppressed recipient, what happens?

(I can do some investigation here if that would be helpful.)

tests/test_mailtrap_backend.py
```diff
@@ -0,0 +1,299 @@
...
+@override_settings(EMAIL_BACKEND="anymail.backends.mailtrap.EmailBackend")
+class MailtrapBackendImproperlyConfiguredTests(AnymailTestMixin, SimpleTestCase):
+    """Test ESP backend without required settings in place"""
+
+    def test_missing_api_token(self):
+        with self.assertRaises(ImproperlyConfigured) as cm:
+            mail.send_mail("Subject", "Message", "from@example.com", ["to@example.com"])
+        errmsg = str(cm.exception)
+        self.assertRegex(errmsg, r"\bMAILTRAP_API_TOKEN\b")
+        self.assertRegex(errmsg, r"\bANYMAIL_MAILTRAP_API_TOKEN\b")
```
medmunds, line 299: Probably also need to test various combinations of settings, both valid and error cases (api_url, testing_enabled, test_inbox_id), to make sure the right api is called or a helpful error is raised.

medmunds (changes requested):
Still looking great! And nice start on the docs.

Handful of feedback on the main implementation.

I glanced through the backend tests and tried to respond to your questions. This is the part of adding an ESP where we typically discover all of the little details they *haven't* documented, so some tests may need adjustment based on that. (The "test_all_options" integration test can be helpful for discovering those.)

(Apologies for not getting feedback on your earlier round sooner—I'm starting to get caught up in the end of the year rush.)

adamwolf:
Hiya folks!

Is there I could help with to help get this ready?

medmunds:
@cahna did you still want to work on this? If not, I'll try to finish it up. (Though realistically, probably not until sometime next month.)

@adamwolf thanks for the offer. There are a couple of things that would be helpful:
- We need to know what Mailtrap's send API response looks like when you try to send to blocked addresses or mixed blocked/not-blocked. (See [comment above](#discussion_r1864691665).)
- If cahna _isn't_ planning to move forward with this PR, you're welcome to adopt it and continue in a new PR. It would probably be helpful to squash the current commits (maintaining [co-author credit](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) for cahna) and continue from there with the unresolved review feedback above.

cahna:
My apologies for ghosting. Work and life has been very busy. Good news is that this has been running in a production environment for 3+ months without any issues (nearing 2 million emails). I want to finish up the integration tests and outstanding comments, but i can't commit to a deadline at this point.

cahna:
I have a question. For this integration test case:

```python3
    def test_all_options(self):
        message = AnymailMessage(
            subject="Anymail Mailtrap all-options integration test",
            body="This is the text body",
            from_email=formataddr(("Test From, with comma", self.from_email)),
            to=[
                "test+to1@anymail.dev",
                'Recipient 2 <test+to2@anymail.dev>',
            ],
            cc=["test+cc1@anymail.dev", "Copy 2 <test+cc2@anymail.dev>"],
            bcc=["test+bcc1@anymail.dev", "Blind Copy 2 <test+bcc2@anymail.dev>"],
            reply_to=[
                '"Reply, with comma" <reply@example.com>',
                "reply2@example.com",
            ],
            headers={"X-Anymail-Test": "value", "X-Anymail-Count": "3"},
            metadata={"meta1": "simple string", "meta2": 2},
            # Mailtrap supports only a single tag/category
            tags=["tag 1"],
            track_clicks=True,
            track_opens=True,
        )
        message.attach("attachment1.txt", "Here is some\ntext for you", "text/plain")
        message.attach("attachment2.csv", "ID,Name\n1,Amy Lina", "text/csv")
        cid = message.attach_inline_image_file(sample_image_path())
        message.attach_alternative(
            "<p><b>HTML:</b> with <a href='http://example.com'>link</a>"
            f"and image: <img src='cid:{cid}'></div>",
            "text/html",
        )

        message.send()
        self.assertEqual(message.anymail_status.status, {"sent"})
        self.assertEqual(
            message.anymail_status.recipients["test+to1@anymail.dev"].status, "sent"
        )
        self.assertEqual(
            message.anymail_status.recipients["test+to2@anymail.dev"].status, "sent"
        )
```

Mailtrap's sandbox receives this raw email:

```
MIME-Version: 1.0
Date: Wed, 13 Aug 2025 19:55:17 +0000
Reply-To: "Reply, with comma" <reply@example.com>, reply2@example.com
X-Anymail-Test: value
X-Anymail-Count: 3
Subject: Anymail Mailtrap all-options integration test
From: "Test From, with comma" <from@anymail.dev>
To: test+to1@anymail.dev, "Recipient 2" <test+to2@anymail.dev>
Cc: test+cc1@anymail.dev, "Copy 2" <test+cc2@anymail.dev>
Content-Type: multipart/mixed;
 boundary=92f2d5e7accc549534b58d936f37be57c201c16cc2ef5adac662920af5e9

--92f2d5e7accc549534b58d936f37be57c201c16cc2ef5adac662920af5e9
Content-Type: multipart/related;
 boundary=270a8fa0ade9a71b202ef006e9c638d823960b3497f12fa7496396bdf2c5

--270a8fa0ade9a71b202ef006e9c638d823960b3497f12fa7496396bdf2c5
Content-Type: multipart/alternative;
 boundary=326c44ce490a35bf5801d95597ed546386d37f621bf4b57c687b5605597a

--326c44ce490a35bf5801d95597ed546386d37f621bf4b57c687b5605597a
Content-Transfer-Encoding: quoted-printable
Content-Type: text/plain; charset=UTF-8

This is the text body
--326c44ce490a35bf5801d95597ed546386d37f621bf4b57c687b5605597a
Content-Transfer-Encoding: quoted-printable
Content-Type: text/html; charset=UTF-8

<p><b>HTML:</b> with <a href=3D'http://example.com'>link</a>and image: <img=
 src=3D'cid:175511491762.9071.11355961105641567026.img@inline'></div>
--326c44ce490a35bf5801d95597ed546386d37f621bf4b57c687b5605597a--

--270a8fa0ade9a71b202ef006e9c638d823960b3497f12fa7496396bdf2c5
Content-Disposition: inline; filename="sample_image.png"
Content-ID: <175511491762.9071.11355961105641567026.img@inline>
Content-Transfer-Encoding: base64
Content-Type: image/png

iVBORw0KGgoAAAANSUhEUgAAACAAAAAgCAYAAABzenr0AAAABHNCSVQICAgIfAhkiAAAAAlwSFlz
AAALEgAACxIB0t1+/AAAABR0RVh0Q3JlYXRpb24gVGltZQAzLzEvMTNoZNRjAAAAHHRFWHRTb2Z0
d2FyZQBBZG9iZSBGaXJld29ya3MgQ1M1cbXjNgAAAZ1JREFUWIXtl7FKA0EQhr+TgIFgo5BXyBUp
fIGksLawUNAXWFFfwCJgBAtfIJFMLXgQn8BSwdpCiPcKAdOIoI2x2Dmyd7kYwXhp9odluX/uZv6d
nZu7DXowxiKZi0IAUHKCvxcsoAIEpST4IawVGb0Hb0BlpcigefACvAAvwAsoTTGGlwwzBAyivLUP
EZrOM10AhGOH2wWugVVlHoAdhJHrPC8DNR0JGsAAQ9mxNzBOMNjS4Qrq69U5EKmf12ywWVsQI4QI
IbCn3Gnmnk7uk1bokfooI7QRDlQIGCdzPwiYh0idtXNs2zq3UqwVEiDcu/R0DVjUnFpItuPSscfA
FXCGSfEAdZ2fVeQ68OjYWwi3ycVvMhABGwgfKXZScHeZ+4c6VzN8FbuYukvOykCs+z8PJ0xqIXYE
d4ALoKlVH2IIgUHWwd/6gNAFPjPcCPvKNTDcYAj1lXzKc7GIRrSZI6yJzcQ+dtV9bD+IkHThBj34
4j9/yYxupaQbXPJLNqsGFgeZ6qwpLP1b4AV4AV5AoKfjpR5OwR6VKwULCAC+AQV4W9Ps4uZQAAAA
AElFTkSuQmCC
--270a8fa0ade9a71b202ef006e9c638d823960b3497f12fa7496396bdf2c5--

--92f2d5e7accc549534b58d936f37be57c201c16cc2ef5adac662920af5e9
Content-Disposition: attachment; filename="attachment1.txt"
Content-Transfer-Encoding: base64
Content-Type: text/plain

SGVyZSBpcyBzb21lCnRleHQgZm9yIHlvdQ==
--92f2d5e7accc549534b58d936f37be57c201c16cc2ef5adac662920af5e9
Content-Disposition: attachment; filename="attachment2.csv"
Content-Transfer-Encoding: base64
Content-Type: text/csv

SUQsTmFtZQoxLEFteSBMaW5h
--92f2d5e7accc549534b58d936f37be57c201c16cc2ef5adac662920af5e9--
```

The `parsed_response` from Mailtrap's API contains:

```json
{'success': True, 'message_ids': ['5038769279']}
```

I suppose this means that all of the `to`, `cc`, and `bcc` addresses should receive the same `AnymailRecipientStatus`? (this means the `zip` code I have doesn't work and needs fixing)

cahna:
For the integration tests of templates, would you please create a template in your mailtrap account?

Needed steps:

1. Go to templates to add a new template with these details:
   image
2. Select "Welcome email" and leave it unchanged:
   image
3. Click "Finish":
   image
4. Copy the template UUID:
   image
5. Add the template UUID to github actions variables as `ANYMAIL_TEST_MAILTRAP_TEMPLATE_UUID`

cahna:
It appears that bcc testing is not available on their free plan:

image

anymail/backends/mailtrap.py
```diff
@@ -0,0 +1,272 @@
...
+            self.data["category"] = tags[0]
+
+    def set_track_clicks(self, *args, **kwargs):
+        """Do nothing. Mailtrap supports this, but it is not configured in the send request."""
+        pass
+
+    def set_track_opens(self, *args, **kwargs):
+        """Do nothing. Mailtrap supports this, but it is not configured in the send request."""
+        pass
+
```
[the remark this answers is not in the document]
cahna: Thanks. I have removed these methods.

anymail/backends/mailtrap.py
cahna: From my integration testing, it seems that there is always exactly one message id returned in `message_ids`. Can you find or think of any ways this would return more than one message_id?

I have updated the code to expect only one message id in the response and associated it with all addresses.

tests/test_mailtrap_backend.py
cahna: I have removed this test, but I have not investigated/tested the cases you mentioned.

medmunds:
> > The `parsed_response` from Mailtrap's API contains:
> > ```json
> > {'success': True, 'message_ids': ['5038769279']}
> > ```
> >   
> > I suppose this means that all of the `to`, `cc`, and `bcc` addresses should receive the same `AnymailRecipientStatus`? (this means the `zip` code I have doesn't work and needs fixing)
> 
> That sounds correct: if there's only a single message id, it applies to all recipients. I'll try to do some live testing to confirm.

Hmm, If I send to two "to" and two "cc" recipients on a *production* domain (`https://send.api.mailtrap.io/api/send` endpoint), I get this response, which seems to mean your zip code is correct:
```json
{
  "success": true,
  "message_ids": [
    "ed1edba0-795c-11f0-0000-f14311a5578e",
    "ed1fc600-795c-11f0-0000-f14311a5578e",
    "ed20fe80-795c-11f0-0000-f14311a5578e",
    "ed21c1d0-795c-11f0-0000-f14311a5578e"
  ]
}
```

But on a *sandbox* domain (`https://sandbox.api.mailtrap.io/api/send/SANDBOX_ID`), I get a single message_id:
```json
{
  "success": true,
  "message_ids": [
    "5040626968"
  ]
}
```

cahna:
>  the 3500 messages/month on the free plan will way more than cover Anymail's test sending

@medmunds Just a heads-up that the "sandbox" is only limited to 100/month

cahna:
> > > The `parsed_response` from Mailtrap's API contains:
> > > ```json
> > > {'success': True, 'message_ids': ['5038769279']}
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I suppose this means that all of the `to`, `cc`, and `bcc` addresses should receive the same `AnymailRecipientStatus`? (this means the `zip` code I have doesn't work and needs fixing)
> > 
> > 
> > That sounds correct: if there's only a single message id, it applies to all recipients. I'll try to do some live testing to confirm.
> 
> Hmm, If I send to two "to" and two "cc" recipients on a _production_ domain (`https://send.api.mailtrap.io/api/send` endpoint), I get this response, which seems to mean your zip code is correct:
> 
> ```json
> {
>   "success": true,
>   "message_ids": [
>     "ed1edba0-795c-11f0-0000-f14311a5578e",
>     "ed1fc600-795c-11f0-0000-f14311a5578e",
>     "ed20fe80-795c-11f0-0000-f14311a5578e",
>     "ed21c1d0-795c-11f0-0000-f14311a5578e"
>   ]
> }
> ```
> 
> But on a _sandbox_ domain (`https://sandbox.api.mailtrap.io/api/send/SANDBOX_ID`), I get a single message_id:
> 
> ```json
> {
>   "success": true,
>   "message_ids": [
>     "5040626968"
>   ]
> }
> ```

Hmm... strange behavior. What do you think would be best to do here? If `self.testing_enabled` expect one ID and use it for all recipients, but if not, use the `zip` code?

I have only been using Anymail to send single-recipient emails, so this is not something I had encountered, yet.

medmunds:
> What do you think would be best to do here? If `self.testing_enabled` expect one ID and use it for all recipients, but if not, use the `zip` code?

That seems reasonable. Maybe check that the length is as expected (1 for sandbox, # recipients for production), just in case they change an API later.

I suspect the difference is that the sandbox records one "message" per API call, however many recipients it has, and the message id is meant to retrieve that single sandbox record. But production email must actually send a separate message to each recipient, and uses individual message ids so you can match a tracking event to its recipient. (Django's mail test outbox has [a similar distinction](https://github.com/anymail/django-anymail/issues/249#issuecomment-939093622).)

medmunds:
> > the 3500 messages/month on the free plan will way more than cover Anymail's test sending
> 
> @medmunds Just a heads-up that the "sandbox" is only limited to 100/month

Oh, hadn't noticed that, thanks.

I think the math is probably still OK. The sandbox is 100 send API calls, with no limit on number of recipients per "send". Production is 3500 messages sent, where each recipient counts as a separate message. (It's the multiple recipients on something like test_all_options that eats through some ESP's smaller quotas. But if we start bumping into the sandbox limit maybe we investigate combining sandbox integration tests.)

While I'm thinking about sandbox vs production:
* I suspect we should have separate (but similar) integration tests for sandbox and production. They're really two different APIs.
* I'm wondering if it would make sense to combine the MAILTRAP_TESTING and MAILTRAP_TEST_INBOX_ID settings? Or really eliminate the separate "testing" one. If test_inbox_id is set, that implies testing_enabled. 🤔

medmunds (dismissed):

medmunds:
@cahna I rebased your original commit and then added a few changes per the discussions above.

I'll try to move this forward over the weekend or early next week. I think the primary remaining work is:
- [ ] Use Mailtrap's batch APIs with Anymail's `merge_data` and `merge_metadata`
- [ ] Make sure the backend tests aren't missing any cases (ones we usually cover with other ESPs, plus some Mailtrap-specific logic around sandbox vs transactional APIs)
- [ ] Update the integration tests to cover both sandbox and transactional APIs
- [ ] Finish the docs

Also, I de-emphasized support for Mailtrap's bulk API in the docs, because I couldn't tell whether it was really the same API payloads/responses as transactional just at a different endpoint. (And didn't want to add to the testing burden if not.)

medmunds:
@cahna if you don't see any problems, I'll plan to merge this in the next few days. Thanks again for the solid foundation.

(I've changed the wording from "test inbox" to "sandbox" to match the terminology Mailtrap seems to be using in their dashboard and docs.)

(Ideally I'd like to find a solution for the rate limit issue on the sandbox integration tests before merging. But I'm not sure that's strictly necessary since the API is identical to the transactional one.)

medmunds:
@cahna thanks again for all the work you put into this. 🎉 

And thanks to Mailtrap for becoming an Anymail sponsor 💚 and to Mailtrap's @yanchuk and @leonid-shevtsov for actively participating in the discussion (and relaxing the limits on the Anymail test account so we can do full integration testing).

PR #415, opened 2025-01-07
Test Django 5.1 on Python 3.12

mgrdcm:
It's supported and the comment suggests it's supposed to be there already.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -10,7 +10,7 @@ envlist =
     docs
     # ... then test all the other supported combinations:
     # Django 5.1: Python 3.10, 3.11, and 3.12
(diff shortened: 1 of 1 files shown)


medmunds:
django51-py312-all is already included on line 9. (You can see that it [was included](https://github.com/anymail/django-anymail/actions/runs/12582246852/job/35067555748) in last Thursday's scheduled run.) This change makes that environment run twice.

The earliest and latest supported environments are broken out at the top of tox.ini so that they run before other tests. Doing that helps catch version-related problems quickly.

mgrdcm:
Gotcha!  Thanks and sorry I missed that!

PR #419, opened 2025-02-17
Postal (docs): Clarify Postal API URL documentation

edgarsj:
POSTAL_API_URL should be the base URL without API paths, as Anymail appends these automatically.

Clarifying as the current name and docs make it ambiguous.

Spent an hour debugging after https://mypostalserver.example.com/api/v1 value didn't work and want to save this for others :)


medmunds:
Sorry that tripped you up, and thanks for the improvement to the docs.

PR #426, opened 2025-04-03
Resend: Avoid installing broken version of svix package

medmunds:
Work around svix/svix-webhooks#1862


medmunds:
(Holding off on merging—looks like Svix is working on a fix.)

medmunds:
Broken version was yanked from PyPI

PR #430, opened 2025-07-22
Add optional webhook signature verification for SendGrid webhooks

blag:
This adds a mixin for the SendGrid webhooks that verifies the signature headers sent when the user has enabled signature verification in the SendGrid webhook console. If not configured, the behavior of the webhooks will not be changed (eg: verification will not be done).

Currently untested and undocumented, just seeing if this change seems reasonable to upstream. Please leave a comment if I'm off base or a 👍 if this looks good so far. Once I get some feedback I'll add tests and documentation.


medmunds:
Thanks for this. I'd normally be very interested, but we've lost the ability to test SendGrid (see #462).

If you're able to perform live testing yourself to verify it, I'd still consider merging it. (Especially since you've made it opt-in so it won't break any existing installations. I typically run a local test project exposed via ngrok or cloudflare tunnel to verify Anymail's webhook signature logic during development.)

I'm a little concerned about the use of hazmat primitives, but would need to review more closely.

blag:
Manually tested against SendGrid itself.

The RTD job seems to have completed successfully if you click through to the actual job, but reporting back to GitHub seems to not have happened as of this comment. Not sure what's up with that or if there's anything I can do to fix it.

This should be ready for review and merge.

Thank you for all of your work on this. It's a lovely third party app.

medmunds:
> sendgrid-python (the official, first-party library developed by SendGrid themselves) uses [ecdsa](https://github.com/sendgrid/sendgrid-python/blob/7eafe1854f42ef2d6863a7288c1c2e73f9aa5c82/requirements.txt#L5)

Oh interesting, sendgrid-python is using a library that specifically says "[should not be used in production settings](https://github.com/tlsfuzzer/python-ecdsa?tab=readme-ov-file#:~:text=NOTE%3A%20This%20library%20should%20not%20be%20used%20in%20production%20settings%2C%20see%20Security%20for%20more%20details.)" and recommends using cryptography instead. Yeah, cryptography is definitely the better choice.

blag:
> For the docs, is there anything that needs to be done in SendGrid's control panel to enable webhook signatures?

Yes.

> Is there an option?

Yes.

> Does it default to on?

No.

The webhook verification defaults to OFF and must be turned on in the SendGrid Mail settings. The titles of the pages I mentioned should direct Anymail users to the correct page on SendGrid. I was a little leery about throwing in links directly to the console page because Twilio/SendGrid's authentication flow has been "unstable" for me so I don't have a huge amount of trust that those URLs won't change. Same story with screenshots. I consciously tried to write my documentation to strike a balance between having to maintain it amidst a challenging environment (read: Twilio broke SendGrid's web auth for me for an entire day after one of their updates) and giving enough breadcrumbs for a reasonably competent developer could find the correct settings.

SendGrid also has their own documentation for their event webhook:
https://www.twilio.com/docs/sendgrid/for-developers/tracking-events/getting-started-event-webhook-security-features#the-signed-event-webhook

medmunds (changes requested):
Thanks, this looks really good. I have a few comments in the implementation, and there are some missing test cases.

We definitely need to make sure we're testing all combinations of:
- SENDGRID_WEBHOOK_KEY set, WEBHOOK_SECRET not
  - with cryptography installed
  - without cryptography installed (error)
- WEBHOOK_SECRET set, SENDGRID_WEBHOOK_KEY not
  - should work without cryptography installed 
- Both set
  - requires cryptography installed; should fail if signature is valid but basic auth header isn't
- Neither set: should warn

(The cryptography-not-installed cases are run by the [`…-none` tox environments](https://github.com/anymail/django-anymail/blob/30c613003f4a8bac0b78d50cabcfd474bf9ea8aa/tox.ini#L53-L56).)

anymail/webhooks/sendgrid.py
```diff
@@ -14,10 +21,97 @@
...
+                ).encode("utf-8"),
+                backend=default_backend(),
+            )
+        super().__init__(**kwargs)
+
+    def validate_request(self, request):
+        # TODO: Remove this call to super(), since this is handled automatically without needing
+        #       superclass chaining
+        # Do basic auth validation first, since it's probably cheaper than signature validation
+        super().validate_request(request)
```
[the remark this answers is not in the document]
blag: 🤦 Oops, good catch. I meant to remove this before pushing and forgot. Will remove and re-test.

tests/test_sendgrid_webhooks.py
```diff
@@ -24,6 +27,81 @@ def call_webhook(self):
     # Actual tests are in WebhookBasicAuthTestCase
 
 
+@tag("sendgrid")
+@ignore_warnings(category=AnymailNotSupportedWarning)
+@unittest.skipUnless(
+    ClientWithSendGridSignature, "Install 'cryptography' to run sendgrid webhook tests"
+)
+class SendGridSignedWebhookSecurityTestCase(WebhookTestCase):
```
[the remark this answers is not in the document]
blag: This got me thinking - I never actually manually tested the inbound webhook. And while I _think_ it would use the exact same verification method, I can't validate that with the SendGrid subuser account that I have access to. So I entirely removed the changes to the Inbound webhook.

Adding support back should be as easy as:

```diff
diff --git a/anymail/webhooks/sendgrid.py b/anymail/webhooks/sendgrid.py
index 239e326..241ff0d 100644
--- a/anymail/webhooks/sendgrid.py
+++ b/anymail/webhooks/sendgrid.py
@@ -234,10 +234,11 @@ class SendGridTrackingWebhookView(SendGridBaseWebhookView):
     }


-class SendGridInboundWebhookView(AnymailBaseWebhookView):
+class SendGridInboundWebhookView(SendGridBaseWebhookView):
     """Handler for SendGrid inbound webhook"""

     esp_name = "SendGrid"
+    webhook_type = "inbound"
     signal = inbound

     def __init__(self, **kwargs):
```

blag:
Expanded tests, and separated out the SendGrid tracking webhook verification key out from a potential Inbound webhook verification key that I _think_ would be a separate key. Can't currently test Inbound webhooks, so this PR doesn't touch that webhook at all.

This should be ready for re-review.

medmunds:
Great! I pushed a few changes. If it looks OK to you, I think it's good to merge.

My changes were mostly minor cleanup, but a few larger ones of note:
- I reworked the setting in SendGridBaseWebhookView slightly to ensure it can also be passed to as_view(), and added a test for that. (We try to [allow as_view() overrides](https://anymail.dev/en/stable/installation/#anymail-settings-reference:~:text=Finally%2C%20for%20complex%20use%20cases%2C%20you%20can%20override%20most%20settings%20on%20a%20per%2Dinstance%20basis%20by%20providing%20keyword%20args%20where%20the%20instance%20is%20initialized%20(e.g.%2C%20in%20a%20get_connection()%20call%20to%20create%20an%20email%20backend%20instance%2C%20or%20in%20a%20View.as_view()%20call%20to%20set%20up%20webhooks%20in%20a%20custom%20urls.py).) for all webhook-related settings.)
- Removed test_successful_webhook_without_signature_or_http_auth which was already covered by SendGridWebhookSecurityTestCase(WebhookBasicAuthTestCase). (And tried to clarify which tests are run with/without cryptography.)
- Expanded the docs a bit, including recommending signature verification for most users.

medmunds (approved):

docs/esps/sendgrid.rst
```diff
@@ -407,13 +453,45 @@ in either Anymail settings or esp_extra as described above.)
...
-   :samp:`https://{random}:{random}@{yoursite.example.com}/anymail/sendgrid/tracking/`
+SendGrid optionally provides webhook signature verification. You have three
+choices for securing the status tracking webhook:
 
-     * *random:random* is an :setting:`ANYMAIL_WEBHOOK_SECRET` shared secret
-     * *yoursite.example.com* is your Django site
+* Use SendGrid's signature verification: follow their
+  `Event Webhook Security Features`_ documentation and set
+  :setting:`SENDGRID_TRACKING_WEBHOOK_VERIFICATION_KEY <ANYMAIL_SENDGRID_TRACKING_WEBHOOK_VERIFICATION_KEY>`
+  (requires the :pypi:`cryptography` package---see :ref:`sendgrid-installation`)
```
blag, line 465:
```suggestion
  (requires the :pypi:`cryptography` package --- see :ref:`sendgrid-installation`)
```

blag, line 465:
```suggestion
  (requires the :pypi:`cryptography` package --- see :ref:`sendgrid-installation`)
```

docs/esps/sendgrid.rst
```diff
@@ -407,13 +453,45 @@ in either Anymail settings or esp_extra as described above.)
...
+
+.. _Event Webhook Security Features:
+    https://www.twilio.com/docs/sendgrid/for-developers/tracking-events/getting-started-event-webhook-security-features#the-signed-event-webhook
+
+
+To configure Anymail status tracking for SendGrid, enter one of these urls
+in your `SendGrid mail settings`_ under "Event Notification" (substituting
+your Django site for *yoursite.example.com*):
+
+*   If are *not* using Anymail's shared webhook secret:
```
blag, line 487:
```suggestion
*   If you are *not* using Anymail's shared webhook secret:
```

pyproject.toml
```diff
@@ -79,7 +79,13 @@ mailjet = []
...
 postmark = []
 resend = ["svix"]
-sendgrid = []
+sendgrid = [
+    # SendGrid optionally uses cryptography for verifying webhook signatures.
+    # Cryptography's wheels are broken on darwin-arm64 before Python 3.9,
+    # and unbuildable on PyPy 3.8 due to PyO3 limitations. Since cpython 3.8
+    # has also passed EOL, just require Python 3.9+ with SendGrid.
+    "cryptography; python_version >= '3.9'"
+]
```
blag, line 88: \*chef's kiss\*

blag (commented):
Two tiny nits but otherwise looks good.

blag:
Really appreciate the cleanup!

blag:
✅ Successfully tested using a live SendGrid account against current `HEAD` for this branch.

docs/esps/sendgrid.rst
medmunds: (I was always told not to put spaces around em dashes, though I gather both styles are common. Anymail's docs don't really have a style guide, but none of the other em dashes have spaces.)

PR #435, opened 2025-08-02
Scaleway: new ESP

hleroy:
Add support for Scaleway Transactional Email backend (but not yet webhooks)


anymail/backends/scaleway.py
```diff
@@ -0,0 +1,163 @@
...
+                "Invalid response from Scaleway API",
+                email_message=message,
+                payload=payload,
+                response=response,
+                backend=self,
+            )
+
+        for email_info in emails:
+            recipient = email_info.get("mail_rcpt")
+            message_id = email_info.get("id")
```
medmunds, line 64: (Note to self: using the response's `"id"` field as Anymail's `message_id` is correct ✔️. Scaleway describes it as the message's "technical id," and it matches the `"email_id"` provided in the tracking [webhook payload](https://www.scaleway.com/en/docs/transactional-email/reference-content/webhook-events-payloads/). Scaleway also provides a `"message_id"` field which is a portion of the *Message-Id* header sent with the email, but that value seems to be useless for webhooks or calling other Scaleway APIs like "get email".)

anymail/backends/scaleway.py
```diff
@@ -0,0 +1,163 @@
...
+    def init_payload(self):
+        self.data = {
+            "to": [],
+            "cc": [],
+            "bcc": [],
+            "additional_headers": [],
+            "attachments": [],
+        }
+        if self.project_id:
+            self.data["project_id"] = self.project_id
```
medmunds, line 103: Is project_id required?

anymail/backends/scaleway.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def set_from_email(self, email):
+        self.data["from"] = {
+            "email": email.addr_spec,
+            "name": email.display_name or None,
+        }
+
+    def set_recipients(self, recipient_type, emails):
+        if emails:
+            self.data[recipient_type] = [
```
medmunds, line 113:
```suggestion
        assert recipient_type in {"to", "cc", "bcc"}
        if emails:
            self.data[recipient_type] = [
```
(Just to double check Anymail's param still matches Scaleway's API payload field name.)

anymail/backends/scaleway.py
```diff
@@ -0,0 +1,163 @@
...
+        return "emails"
+
+    def init_payload(self):
+        self.data = {
+            "to": [],
+            "cc": [],
+            "bcc": [],
+            "additional_headers": [],
+            "attachments": [],
+        }
```
medmunds, line 101: I would suggest *not* initializing the empty fields here. That avoids needing to delete them if unused later. (Your other methods all create the fields as needed.)
```suggestion
        self.data = {}
```

anymail/backends/scaleway.py
```diff
@@ -0,0 +1,163 @@
...
+            del self.data["to"]
+        if not self.data.get("cc"):
+            del self.data["cc"]
+        if not self.data.get("bcc"):
+            del self.data["bcc"]
+        if not self.data.get("attachments"):
+            del self.data["attachments"]
+        if not self.data.get("additional_headers"):
+            del self.data["additional_headers"]
+        return self.serialize_json(self.data)
```
medmunds, line 163:
```suggestion
    def serialize_data(self):
        return self.serialize_json(self.data)
```

(If you skip creating the empty lists in init_payload.)

docs/esps/esp-feature-matrix.csv
```diff
@@ -1,21 +1,21 @@
...
-:attr:`~AnymailMessage.merge_data`,Yes [#caveats]_,Yes,Yes,Yes,Yes,Yes,No,Yes,No,Yes,Yes,Yes
-:attr:`~AnymailMessage.merge_global_data`,Yes [#caveats]_,Yes,Yes,Yes,Yes,Yes,No,Yes,No,Yes,Yes,Yes
-.. rubric:: :ref:`Status <esp-send-status>` and :ref:`event tracking <event-tracking>`,,,,,,,,,,,,
-:attr:`~AnymailMessage.anymail_status`,Yes,Yes,Yes,Yes,Yes,Yes,Yes,Yes,Yes,Yes,Yes,Yes
-:class:`~anymail.signals.AnymailTrackingEvent` from webhooks,Yes,Yes,Yes,Yes,Yes,Yes,Yes,Yes,Yes,Yes,Yes,Yes
-.. rubric:: :ref:`Inbound handling <inbound>`,,,,,,,,,,,,
-:class:`~anymail.signals.AnymailInboundEvent` from webhooks,Yes,Yes,Yes,Yes,Yes,Yes,Yes,Yes,No,Yes,Yes,No
+Email Service Provider,:ref:`amazon-ses-backend`,:ref:`brevo-backend`,:ref:`mailersend-backend`,:ref:`mailgun-backend`,:ref:`mailjet-backend`,:ref:`mandrill-backend`,:ref:`postal-backend`,:ref:`postmark-backend`,:ref:`resend-backend`,:ref:`scaleway-backend`,:ref:`sendgrid-backend`,:ref:`sparkpost-backend`,:ref:`unisender-go-backend`
+Anymail support status [#support-status]_,Full,Full,Full,Full,Full,Limited,Limited,Full,Full,Limited,**Unsupported**,Full,Full
```
medmunds, line 2: I *think* we'll be able to mark Scaleway as "Full" support, because we'll be able to run regular integration tests.

(I've set up a test account for Anymail, and Scaleway's free 300 messages/month should be plenty for integration testing.)

docs/esps/scaleway.rst
```diff
@@ -0,0 +1,215 @@
...
+
+**No delayed sending**
+  Scaleway does not support :attr:`~anymail.message.AnymailMessage.send_at`.
+
+**No tags support**
+  Scaleway does not support :attr:`~anymail.message.AnymailMessage.tags`.
+
+**No metadata support**
+  Scaleway does not support :attr:`~anymail.message.AnymailMessage.metadata` or
+  :attr:`~anymail.message.AnymailMessage.merge_metadata` features.
```
medmunds, line 177: We *could* support metadata and tags by sending them in custom message headers. The message headers seem to be included in the webhook payload, so could be retrieved by Anymail's tracking webhook.

(We do this for Resend. If you're interested, take a look at anymail/backends/resend.py.)

pyproject.toml
```diff
@@ -33,6 +33,7 @@ keywords = [
     "SendGrid", "Twilio",
     "SparkPost", "Bird",
     "Unisender Go",
+    "Scaleway",
```
medmunds, line 36: Please keep these in alphabetical order.

pyproject.toml
```diff
@@ -83,6 +84,7 @@ sendgrid = []
 sendinblue = []
 sparkpost = []
 unisender-go = []
+scaleway = []
```
medmunds, line 87: Please keep these in alphabetical order. (I'm not sure why postal is at the end.)

tests/test_scaleway_backend.py
```diff
@@ -0,0 +1,109 @@
...
+@override_settings(
+    EMAIL_BACKEND="anymail.backends.scaleway.EmailBackend",
+    ANYMAIL={"SCALEWAY_API_KEY": "test_api_key"},
+)
+class ScalewayBackendMockAPITestCase(RequestsBackendMockAPITestCase):
+    DEFAULT_RAW_RESPONSE = b"""{
+        "emails": [
+            {
+                "id": "b7c6a8c1-19a4-49d8-9293-41835462ce81",
+                "message_id": "<b7c6a8c1-19a4-49d8-9293-41835462ce81@example.com>",
```
medmunds, line 19: Just FYI, in my quick testing, I got two different UUIDs for id and message_id, and the message_id was not formatted as an email *Message-Id*:

```
      "id": "c4a471a9-d195-4dc1-b55e-69cd3dece155",
      "message_id": "1098e9fd-f27f-45f3-9cbe-487ec1dccaec",
```

But the email it sent had this *Message-Id* header, which includes the API response's message_id:

```
Message-Id: <1098e9fd-f27f-45f3-9cbe-487ec1dccaec@phan-van6.scw-tem.cloud>
```

tests/test_scaleway_backend.py
```diff
@@ -0,0 +1,109 @@
...
+
+@tag("scaleway")
+class ScalewayBackendAnymailFeatureTests(ScalewayBackendMockAPITestCase):
+    def test_extra_headers(self):
+        self.message.extra_headers = {"X-Custom": "string", "X-Num": 123}
+        self.message.send()
+        data = self.get_api_call_json()
+        headers = {h["key"]: h["value"] for h in data["additional_headers"]}
+        self.assertEqual(headers["X-Custom"], "string")
+        self.assertEqual(headers["X-Num"], 123)
```
medmunds, line 94: I'd suggest *not* composing a headers dict here, but instead asserting that the generated json is the correct structure:
```suggestion
        data = self.get_api_call_json()
        self.assertCountEqual(
            data["additional_headers"],
            [
                {"key": "X-Custom", "value": "string"},
                {"key": "X-Num", "value": 123},
            ],
        )
```

(assertCountEqual works even if the additional_headers list is serialized in a different order.)

Same thing in the next test, too.

tests/test_scaleway_backend.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def test_reply_to(self):
+        self.message.reply_to = ["reply@example.com"]
+        self.message.send()
+        data = self.get_api_call_json()
+        headers = {h["key"]: h["value"] for h in data["additional_headers"]}
+        self.assertEqual(headers["Reply-To"], "reply@example.com")
+
```
medmunds, line 102: Since reply_to uses the same payload field as extra_headers, we should also have a test that uses both and makes sure additional_headers ends up with both of them.

tests/test_scaleway_backend.py
```diff
@@ -0,0 +1,109 @@
...
+
+@tag("scaleway")
+@override_settings(EMAIL_BACKEND="anymail.backends.scaleway.EmailBackend")
+class ScalewayBackendImproperlyConfiguredTests(SimpleTestCase):
+    def test_missing_api_key(self):
+        with self.assertRaises(AnymailConfigurationError):
+            mail.send_mail("Subject", "Message", "from@example.com", ["to@example.com"])
```
medmunds, line 109: If project_id is also required, we should test that.

anymail/backends/scaleway.py
```diff
@@ -0,0 +1,163 @@
...
+    def set_text_body(self, body):
+        if body:
+            self.data["text"] = body
+
+    def set_html_body(self, html):
+        if html:
+            self.data["html"] = html
+
+    def add_attachment(self, attachment):
+        self.data.setdefault("attachments", []).append(
```
medmunds, line 131: It looks like Scaleway's API doesn't have any way to specify inline images, so that counts as an "unsupported feature":

```suggestion
        if attachment.inline:
            self.unsupported_feature("inline attachments")
        self.data.setdefault("attachments", []).append(
```

(If the user is ignoring unsupported feature errors, this will fall through and handle the inline attachment as a regular attachment.)

medmunds:
> Regarding webhooks, I started coding the implementation. Then when comparing with other ESPs, I was disappointed to notice that Scaleway doesn't offer yet some signing or authentication feature on the webhook payload.

Fwiw, many (most?) ESPs don't sign their webhooks, and instead support a shared secret transmitted as HTTP basic auth for authentication. It seems like Scaleway allows including basic auth credentials in the webhook subscription url, so that's a reasonable option. (It's effectively the same as using a bearer token, which is how Scaleway authenticates calls to their email API. Their Subscriptions control panel unfortunately displays the credentials in plaintext, which isn't ideal, but if someone gains access to the control panel there are probably bigger things to worry about.)

> I'd rather get this PR merged first, then prepare the webhook code / tests, and wait for Scaleway to announce the end of beta.

👍 

Appreciate the quick update. At a glance, everything looks good. I'll try to take a closer look and get it merged later this week.

hleroy:
Hi Mike - does it look good for merging?

medmunds:
@hleroy apologies for the delay.

I made two functional changes:
- Allowed the SCALEWAY_REGION and SCALEWAY_API_URL settings to be used in combination
- Implemented esp_extra in the Scaleway backend

I also added some test cases borrowed from some other ESPs (which is how I noticed esp_extra wasn't implemented), and updated the docs a bit for consistency with the other ESPs.

I think this is ready to merge. Thanks again for all your work on this!

PR #455, opened 2025-12-18
CI/CD: reduce permissions to GitHub workflows

medmunds:
Minimize permissions available to workflow jobs.


medmunds:
Merged via d6e837457a878088d0b96acea44af96b66244925

PR #458, opened 2026-03-09
check typing with mypy

terencehonles:
This picks up stalled PR #394. I'm not sure if squashed PRs are the default, but it makes sense to squash all this work to get rid of the earlier churn, but keep the co-authorship.

### DIFF
--- a/anymail/backends/amazon_ses.py
+++ b/anymail/backends/amazon_ses.py
@@ -1,6 +1,9 @@
+from __future__ import annotations
+
 import email.charset
 import email.encoders
 import email.policy
+from typing import Any
 
 from requests.structures import CaseInsensitiveDict
 
@@ -165,6 +168,7 @@ class AmazonSESV2SendEmailPayload(AmazonSESBasePayload):
         try:
             for field in address_fields:
                 addresses = getattr(self.message, field)
+                idna_encoded_addresses: str | list[str]
                 idna_encoded_addresses = [
                     address.format(idna_encode=self.backend.idna_encode)
                     for address in parse_address_list(addresses, field)
@@ -364,7 +368,7 @@ class AmazonSESV2SendBulkEmailPayload(AmazonSESBasePayload):
     def init_payload(self):
         super().init_payload()
         # late-bind in finalize_payload:
-        self.recipients = {"to": [], "cc": [], "bcc": []}
+        self.recipients = dict[str, list[Any]](to=[], cc=[], bcc=[])
         self.merge_data = {}
         self.headers = {}
         self.merge_headers = {}
--- a/anymail/backends/amazon_ses.py
+++ b/anymail/backends/amazon_ses.py
@@ -435,7 +435,7 @@ class AmazonSESV2SendBulkEmailPayload(AmazonSESBasePayload):
     def parse_recipient_status(self, response):
         try:
             results = response["BulkEmailEntryResults"]
-            ses_status_set = set(result["Status"] for result in results)
+            ses_status_set = {result["Status"] for result in results}
             anymail_statuses = [
                 AnymailRecipientStatus(
                     message_id=result.get("MessageId", None),
--- a/anymail/backends/base_requests.py
+++ b/anymail/backends/base_requests.py
@@ -150,7 +150,7 @@ class AnymailRequestsBackend(AnymailBaseBackend):
                 method=request.method,
                 url=request.url,
                 headers="".join(
-                    "{header}: {value}\n".format(header=header, value=value)
+                    f"{header}: {value}\n"
                     for (header, value) in request.headers.items()
                 ),
             )
@@ -168,7 +168,7 @@ class AnymailRequestsBackend(AnymailBaseBackend):
                 status=response.status_code,
                 reason=response.reason,
                 headers="".join(
-                    "{header}: {value}\n".format(header=header, value=value)
+                    f"{header}: {value}\n"
                     for (header, value) in response.headers.items()
                 ),
                 body=response.text,  # Let Requests decode body content for us
--- a/anymail/backends/base_requests.py
+++ b/anymail/backends/base_requests.py
@@ -1,3 +1,6 @@
+from __future__ import annotations
+
+from typing import cast
 from urllib.parse import urljoin
 
 import requests
@@ -10,6 +13,8 @@ from .base import AnymailBaseBackend, BasePayload
 
 
 class AnymailRequestsBackend(AnymailBaseBackend):
+    session: requests.Session | None
+
     """
     Base Anymail email backend for ESPs that use an HTTP API via requests
     """
@@ -73,7 +78,7 @@ class AnymailRequestsBackend(AnymailBaseBackend):
         session.headers["User-Agent"] = "django-anymail/{version}-{esp} {orig}".format(
             esp=self.esp_name.lower(),
             version=__version__,
-            orig=session.headers.get("User-Agent", ""),
+            orig=cast(str, session.headers.get("User-Agent", "")),
         )
 
         if self.debug_api_requests:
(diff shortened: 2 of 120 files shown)


anymail/backends/amazon_ses.py
```diff
@@ -1,6 +1,9 @@
+from __future__ import annotations
```
terencehonles, line 1: I've added this `future` import via isort (and updated the config) in order to use the most up to date type definitions and not worry about cases where `ImportError` was caught and a lazy error would cause a runtime issue. This brings all supported versions to how Python 3.14 behaves.

This is only a behavioral change if the types definitions were already used at runtime and this changes how the references need to be looked up in certain cases, but since the types were not used this way I believe this is OK.

anymail/backends/amazon_ses.py
```diff
@@ -364,7 +368,7 @@ class AmazonSESV2SendBulkEmailPayload(AmazonSESBasePayload):
     def init_payload(self):
         super().init_payload()
         # late-bind in finalize_payload:
-        self.recipients = {"to": [], "cc": [], "bcc": []}
+        self.recipients: dict[str, list[Any]] = {"to": [], "cc": [], "bcc": []}
```
terencehonles: I'm leaving narrowing `Any` to the appropriate type as a follow up PR, since I wanted mypy to pass without disabling too many strictness flags, but I didn't want to figure out what was the most narrow type definition would be.

anymail/backends/base_requests.py
```diff
@@ -73,7 +78,7 @@ def create_session(self):
         session.headers["User-Agent"] = "django-anymail/{version}-{esp} {orig}".format(
             esp=self.esp_name.lower(),
             version=__version__,
-            orig=session.headers.get("User-Agent", ""),
+            orig=cast(str, session.headers.get("User-Agent", "")),
```
terencehonles, line 81: I assume the UA is generally a string, the alternative would be to use a byte string and never decode the existing UA and just encode all the other parameters since `bytes` doesn't have a `.format` but it can interpolate.

anymail/backends/brevo.py
```diff
@@ -103,15 +107,15 @@ def __init__(self, message, defaults, backend, *args, **kwargs):
         http_headers["api-key"] = backend.api_key
         http_headers["Content-Type"] = "application/json"
 
-        super().__init__(
+        super().__init__(  # type: ignore[misc]
```
terencehonles, line 110: I just squelched this error, but technically since `*args` is unbounded `headers` can come as a positional argument _or_ a keyword argument. Moving `RequestsPayload` to use keyword only arguments will fix this issue, but is a breaking change.

anymail/webhooks/unisender_go.py
```diff
@@ -71,10 +71,10 @@ def get(
         # Unisender Go verifies the webhook with a GET request at configuration time
         return HttpResponse()
 
-    def parse_json_body(self, request: HttpRequest) -> dict | list | None:
+    def parse_json_body(self, request: HttpRequest) -> dict:
```
terencehonles, line 74: This looks like it was a bad type definition because below it was only used as a dict and would break if it were a list or `None`.

anymail/inbound.py
```diff
@@ -227,7 +230,7 @@ def as_uploaded_file(self):
         name = self.get_filename()
         content_type = self.get_content_type()
         content = self.get_content_bytes()
-        return SimpleUploadedFile(name, content, content_type)
+        return SimpleUploadedFile(name, content, content_type)  # type: ignore[arg-type]
```
terencehonles, line 233: name may be `None` here, but I don't know if the Django stubs type definition is wrong by requiring non `None`, if `get_filename()` might actually not be none, or if this code needs to change.

anymail/inbound.py
```diff
@@ -299,7 +302,7 @@ def construct(
                 raw_headers, headersonly=True
             )
             # headersonly forces an empty string payload, which breaks things later:
-            msg.set_payload(None)
+            msg.set_payload(None)  # type: ignore[call-overload]
```
terencehonles, line 305: The Python typeshed seems to be incorrect if this is valid, but I didn't look into the source to see what happens here, but I'm going to assume this is valid.

anymail/utils.py
```diff
@@ -24,20 +28,49 @@
...
     # Django >= 7.0
-    MIMEMixin = None
+    MIMEMixin = None  # type: ignore[assignment,misc]
 
 from .exceptions import AnymailConfigurationError, AnymailInvalidAddress
 
 BASIC_NUMERIC_TYPES = (int, float)
 
 
-UNSET = type("UNSET", (object,), {})  # Used as non-None default value
```
terencehonles, line 34: This is basically a funny way to write `class UNSET: pass`, which the latter `mypy` can actually use since it's a concrete type, but this is used as a singleton which mypy doesn't like for `is` checks so I've replaced this implementation with an implementation I use for the same functionality including the type that mypy should use.

anymail/utils.py
```diff
@@ -570,7 +607,7 @@ def content_bytes(self):
         """Content as bytes, using original charset"""
         content = self.content
         if isinstance(content, str):
-            content = content.encode(self.charset)
+            content = content.encode(self.charset or settings.DEFAULT_CHARSET)
```
terencehonles, line 601: `charset` seems like it's only guaranteed to be non `None` if it's a "text/*" mime type, I assume this is coming from the user, so if for some reason the charset isn't defined the default makes sense, right?

anymail/utils.py
```diff
@@ -868,6 +905,8 @@ def parse_rfc2822date(s):
 
 
 class CaseInsensitiveCasePreservingDict(CaseInsensitiveDict):
+    _store: OrderedDict
```
terencehonles, line 899: typeshed doesn't define `_store` (trying to keep it private)

anymail/utils.py
terencehonles: It looks like this needs to be pickle safe, so I've used the `TYPE_CHECKING` definition which is more like the original definition where there is a class defined, but this still creates a singleton of that class.

terencehonles (commented):
Some comments about the changes I've made

terencehonles:
@medmunds friendly ping (just making sure you also saw this PR)

medmunds:
Thanks for picking this up! (And of course thanks to @YPCrumble for the earlier work.)

I won't have time to get into details before mid-next-week at the earliest. In the meantime:
- I've created a "next" branch that drops older Python and Django versions, from your other PR, so typing can start with Python 3.10. Suggest rebasing this onto that. (I've already updated the target branch for this PR from main to next.)
- Yes, please squash the earlier commits so we're starting from a simpler diff. You can use Co-Authored-By to preserve YPCrumble's credit.
- I'll glance through your comments and see if I have any quick answers.

anymail/backends/brevo.py
medmunds: All of the payloads can be considered internal types.

fwiw, in other projects I'm starting to drop `*args` and `**kwargs` where they're not absolutely necessary. But maybe start with this change, then improve the API as a follow-on commit.

(btw, this is likely to be part of a breaking change release anyway. We might just add a blanket "internals have been reworked" changelog item.)

anymail/inbound.py
medmunds: There were definitely some significant errors in the Python typeshed around mail last year (when I was looking at this for Django), particularly in legacy APIs like set_payload. I'm not sure the current status.

anymail/utils.py
medmunds: I mean, the right thing to do here is a Sentinel, but that keeps getting deferred.

terencehonles:
I figured we might just target `main` with this and use GitHub's squash feature, but I can squash and retarget the PR

anymail/inbound.py
terencehonles: It looks like this is threading a needle which does work according to the source, but it's unlikely a PR will be accepted to change this. Looking into this more should this be calling the [`clear_content` method](https://github.com/python/cpython/blob/7ba9580f00d3ae2c79cb43f5663725e6d786ef7c/Lib/email/message.py#L1206-L1209) instead? That will strip some headers, but I assume any `content-*` headers aren't valid without content.

PR #459, opened 2026-03-09
drop Django 4.x and EoL Python versions it supported & run pyugrade

terencehonles:


terencehonles:
Django 4.x is pretty old, but this can also wait until 6.1 is out if that is preferred. I've already uninstalled such old versions of Python so it's a bit harder to test PRs without pushing it to the CI (or reinstalling them / using docker / etc)

medmunds:
Thanks for this—I'm very much looking forward to dropping those old Python versions.

[Django 4.2 is still in LTS](https://www.djangoproject.com/download/#:~:text=4.2%20LTS,April%202026) through sometime next month, so we can't actually drop it until then. (We *could* drop Django 4.0 and 4.1 and the EOL Pythons now, but at this point I'm not sure it's worth the hassle to split this PR up. It'll be next month before you know it.)

terencehonles:
Yeah, this can just hang around until then. I just figured I'd create it when working on #458

medmunds:
I'm going to merge this into a new "next" branch, and #458 can work from that. Once Django 4.2 is EoL, we'll replace main with next.

PR #462, opened 2026-03-23
Add Resend inbound handling.

btimby:
Added inbound handling for Resend.

https://resend.com/docs/dashboard/receiving/introduction

### DIFF
--- a/docs/esps/resend.rst
+++ b/docs/esps/resend.rst
@@ -72,10 +72,12 @@ nor ``ANYMAIL_RESEND_API_KEY`` is set.
 
 .. rubric:: RESEND_SIGNING_SECRET
 
-The Resend webhook signing secret used to verify webhook posts.
+The Resend webhook signing secret used to verify *tracking* webhook posts.
 Recommended if you are using activity tracking, otherwise not necessary.
 (This is separate from Anymail's
-:setting:`WEBHOOK_SECRET <ANYMAIL_WEBHOOK_SECRET>` setting.)
+:setting:`WEBHOOK_SECRET <ANYMAIL_WEBHOOK_SECRET>` setting and the
+:setting:`RESEND_INBOUND_SECRET <ANYMAIL_RESEND_INBOUND_SECRET>` used for
+securing *inbound* ``email.received`` webhook posts.)
 
 Find this in your Resend `Webhooks settings`_: after adding
 a webhook, click into its management page and look for "signing secret"
@@ -86,6 +88,30 @@ near the top.
       ANYMAIL = {
           ...
           "RESEND_SIGNING_SECRET": "whsec_...",
+      }
+
+If you provide this setting, the svix package is required.
+See :ref:`resend-installation` above.
+
+
+.. setting:: ANYMAIL_RESEND_INBOUND_SECRET
+
+.. rubric:: RESEND_INBOUND_SECRET
+
+The Resend webhook signing secret used to verify *inbound* ``email.received``
+webhook posts. Recommended if you are using inbound email, otherwise not
+necessary. (This is separate from Anymail's :setting:`WEBHOOK_SECRET
+<ANYMAIL_WEBHOOK_SECRET>` setting.)
+
+Find this in your Resend `Webhooks settings`_: after adding an inbound
+webhook, click into its management page and look for "signing secret"
+near the top.
+
+  .. code-block:: python
+
+      ANYMAIL = {
+          ...
+          "RESEND_INBOUND_SECRET": "whsec_...",
       }
 
 If you provide this setting, the svix package is required.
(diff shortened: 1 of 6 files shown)


btimby:
Sorry for the noise, I decided to update the documentation as well. I don't plan to push anything else unless I find an error or you request changes. I did test this locally with resend and I am receiving emails, using a secret and signing key.

docs/esps/resend.rst
```diff
@@ -384,7 +384,46 @@ you can retrieve them in your tracking signal receiver like this:
...
+    include the *random:random* shared secret in the URL:
+
+    :samp:`https://{random}:{random}@{yoursite.example.com}/resend/inbound/`
+
+*   For "Events to listen", select 'email.received'.
+
+*   Click the "Add" button.
+
+Then, if you are using Resend's webhook signature validation (with svix),
+add the webhook signing secret to your Anymail settings:
```
medmunds, line 410: I would maybe note that this is the same `RESEND_SIGNING_SECRET` used for tracking webhooks, so if you're using both you only need to configure it once.

tests/test_resend_inbound.py
```diff
@@ -0,0 +1,381 @@
...
+            "--boundary\r\n"
+            "Content-Type: text/plain\r\n"
+            "\r\n"
+            "Hello, world!\r\n"
+            "--boundary\r\n"
+            "Content-Type: text/html\r\n"
+            "\r\n"
+            "<p>Hello, world!</p>\r\n"
+            "--boundary--\r\n"
+        )
```
medmunds: This is a style choice, but other Anymail tests typically use `dedent("""\ ... """).replace("\n", "\r\n")` (with a single triple-quoted string) to make inlined mime a little more readable: [example](https://github.com/anymail/django-anymail/blob/e0f50c2550a1f42e287ab63ca0046ba200e77353/tests/test_amazon_ses_inbound.py#L40-L67).

medmunds (commented):
@btimby this looks great. And I appreciate that you added the docs. Thanks!

I have two (very minor) suggestions.

medmunds:
Closes #464.

docs/esps/resend.rst
btimby: Hmm, I did a quick test, if I add a second webhook URL in resend.com, it is assigned a DIFFERENT signing secret. So if you use both (because the events triggering the webhooks differ) then you actually need TWO signing secrets.

tests/test_resend_inbound.py
btimby: OK, I can follow this style.

docs/esps/resend.rst
medmunds: Ah, I was wondering about that. We'll need to use two different settings, then, to allow separate secrets for when the tracking and inbound webhooks are used together. The MailerSend webhooks [do the same thing](https://anymail.dev/en/stable/esps/mailersend/#inbound-routing:~:text=After%20you%20have%20saved%20the%20inbound,be%20sure%20to%20include%20both%20settings.). 

I'd suggest `RESEND_INBOUND_SECRET` (paralleling what we did for MailerSend). You *should* be able to make that work by setting class attr `_secret_setting_name = "inbound_secret"` in `ResendInboundWebhookView`. (There's already support code in `SvixWebhookValidationMixin`.)

btimby: OK, I believe I have covered this by setting the secret setting name attribute on the inbound view. I also updated the docs.

medmunds: This looks right to me.

I'm concerned that no tests failed after you made this change, which makes me think the tests aren't covering signature verification for the inbound webhook.

I'll pull a copy locally over the weekend and try to remember how svix is tested for the tracking webhook.

btimby: Yes, I don't think signature is being tested. Did you have a chance to look into this or do you want me to look at adding signature checking to these tests?

medmunds: Sorry, I haven't had a chance to look at it yet, other than to glance at test_resend_webhooks.py and remember that testing all the combinations of Anymail's shared WEBHOOK_SECRET, Resend's signing key, and svix installed or not got complicated.

If you have a chance to look at it that would be great. The `ResendWebhookSettingsTestCase` and `ResendWebhookSecurityTestCase` tests might have some clues. (svix installed/not is covered by the tox matrix with the none/resend factors in `django60-py314-{none,amazon_ses,postal,resend,sendgrid}`.)

Thanks!

medmunds: OK, I added the signature tests and rebased to main. I'll let the tests finish and then I think this is ready for merge.

Thanks again for implementing this!

PR #468, opened 2026-04-18
Inbound: sanitize CRNL in ESP-parsed attachment filenames

### LINKED ISSUE #465: Postmark webhook fails if any attachment has an invalid name in the Content-Type header
Some of our inbound Postmark events fail completely when an attachment contains a CR/LF sequence in the `name` part of its `Content-Type` header (due to validation further down the stack in Python's `email.policy` `header_store_parse` function).

Ideally, Anymail would either sanitize the value (to accommodate `email.policy`) or allow individual `construct_attachment` calls to fail without failing to construct the message as a whole.

Here's the stack we're seeing:

https://github.com/anymail/django-anymail/blob/e0f50c2550a1f42e287ab63ca0046ba200e77353/anymail/webhooks/postmark.py#L192-L202

calls `construct_attachment`:

https://github.com/anymail/django-anymail/blob/e0f50c2550a1f42e287ab63ca0046ba200e77353/anymail/inbound.py#L396

calls `set_param` with a resolved Content-Type value like `"image/png; name="Outlook-Icon\n\nDesc.png""` as `ctype`:

```python
# email/message.py in set_param at line 794

        if ctype != self.get(header):
            if replace:
                self.replace_header(header, ctype)
            else:
                del self[header]
                self[header] = ctype
```

calls `__setitem__` with that same value as `val`:

```python
# email/message.py in __setitem__ at line 449

        self._headers.append(self.policy.header_store_parse(name, val))
```

calls `header_store_parse` with that same value as `val`:

```python
# email/policy.py in header_store_parse at line 153

        if isinstance(value, str) and len(value.splitlines())>1:
            # XXX this error message isn't quite right when we use splitlines
            # (see issue 22233), but I'm not sure what should happen here.
            raise ValueError("Header values may not contain linefeed "
                             "or carriage return characters")
```

medmunds:
Replace `\r` and `\n` in inbound attachment filenames with space.

Fixes #465.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -41,6 +41,14 @@ Features
 
 * **Resend:** Add support for inbound email. (Thanks to `@btimby`_.)
 
+Fixes
+~~~~~
+
+* **Inbound:** Convert carriage return and line feed characters in attachment
+  filenames to spaces for ESPs that parse them from invalid headers (e.g.,
+  Postmark). (Thanks to `@BHSPitMonkey`_ for reporting the issue and providing
+  an example.)
+
 
 v14.0
 -----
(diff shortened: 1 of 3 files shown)


medmunds:
Merged in 8ad9d6e

PR #475, opened 2026-06-18
Update URLs to Mailjet's documentation

mgrdcm:
Chose the v3.1 version in some cases because that's what `anymail` is using now.

Looks like their docs changed from being under `/guides/` to `/email/guides/` at some point.


medmunds:
Thanks for this. (Keeping up with link rot is a never-ending task.)

mgrdcm:
> Thanks for this. (Keeping up with link rot is a never-ending task.)

You're welcome!  I'm still struggling to get verified commits working so I see merging is blocked for now.

medmunds:
> I'm still struggling to get verified commits working so I see merging is blocked for now.

If you get them working, that's great, otherwise I can work around it to merge.

PR #479, opened 2026-07-25
fix(postmark): report unsupported feature for merge_data without template_id

### LINKED ISSUE #470: Postmark merge_data ignored without template_id
Hi,
I am migrating to Postmark from Sparkpost, and trying to understand the merge behaviour. 

For context, I am sending a HTML email (not ESP template).
When setting merge_data on a message, I can see that TemplateModel is being set. However from the API docs it seems that this is only supported on the new bulk send API, and the template send API endpoints, not the normal batch send API. 

When sending an email I can see that the template variables have not been processed (currently looking at emails in the Postmark sandbox).

Just wondering if there is something I am missing, or is this not a supported use case, and if so, should a Not Supported error be thrown in this circumstance?

Sanjays2402:
Fixes #470

Postmark's batch send API (`/email/batch`) ignores `TemplateModel`, so per-recipient `merge_data` sent without a `template_id` was silently dropped and recipients got the un-substituted body. `serialize_data` now raises `AnymailUnsupportedFeature` in that case, as suggested on the issue. `merge_data={}` still forces batch sending without a template.

New `test_merge_data_without_template_unsupported` fails without the backend change and passes with it; the rest of `tests.test_postmark_backend` (60 tests) is green.

### DIFF
--- a/anymail/backends/postmark.py
+++ b/anymail/backends/postmark.py
@@ -222,11 +222,7 @@ class PostmarkPayload(RequestsPayload):
 
     def get_api_endpoint(self):
         batch_send = self.is_batch()
-        if (
-            "TemplateAlias" in self.data
-            or "TemplateId" in self.data
-            or "TemplateModel" in self.data
-        ):
+        if "TemplateAlias" in self.data or "TemplateId" in self.data:
             if batch_send:
                 return "email/batchWithTemplates"
             else:
@@ -234,6 +230,11 @@ class PostmarkPayload(RequestsPayload):
                 # (Typo?)
                 return "email/withTemplate/"
         else:
+            # Postmark ignores TemplateModel unless using a template API.
+            if "TemplateModel" in self.data:
+                self.unsupported_feature("merge_global_data without template_id")
+            if self.merge_data:
+                self.unsupported_feature("merge_data without template_id")
             if batch_send:
                 return "email/batch"
             else:
--- a/tests/test_postmark_backend.py
+++ b/tests/test_postmark_backend.py
@@ -623,6 +623,39 @@ class PostmarkBackendAnymailFeatureTests(PostmarkBackendMockAPITestCase):
             "e2ecbbfc-fe12-463d-b933-9fe22915106d",
         )
 
+    def test_merge_data_without_template_unsupported(self):
+        # Postmark's batch send API ignores TemplateModel, so merge_data
+        # content can't be applied without an ESP template. (#470)
+        message = AnymailMessage(
+            from_email="from@example.com",
+            to=["alice@example.com", "Bob <bob@example.com>"],
+            subject="Test batch send",
+            body="Hello {{name}}",
+            merge_data={
+                "alice@example.com": {"name": "Alice"},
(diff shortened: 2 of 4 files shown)


tests/test_postmark_backend.py
```diff
@@ -623,6 +623,24 @@ def test_merge_data_no_template(self):
             "e2ecbbfc-fe12-463d-b933-9fe22915106d",
         )
 
+    def test_merge_data_without_template_unsupported(self):
+        # Postmark's batch send API ignores TemplateModel, so merge_data
```
medmunds, line 627: We'll also need a test for `merge_global_data`.

anymail/backends/postmark.py
```diff
@@ -251,6 +251,10 @@ def serialize_data(self):
         elif api_endpoint == "email/batchWithTemplates":
             data = {"Messages": [self.data_for_recipient(to) for to in self.to_emails]}
         elif api_endpoint == "email/batch":
+            if self.merge_data:
+                # Postmark's batch send API ignores TemplateModel. Merge data
+                # is only applied by the template APIs.
+                self.unsupported_feature("merge_data without template_id")
```
medmunds: I think it might be clearer to handle this in `get_api_endpoint()`: if we need to send TemplateModel without a TemplateId or TemplateAlias, then there's no suitable API endpoint available. (That also gives us a place to catch `merge_global_data`, which uses TemplateModel but might be either the `email` or `email/batch` endpoint.

Something like this:

```py
    def get_api_endpoint(self):
        batch_send = self.is_batch()
        if "TemplateAlias" in self.data or "TemplateId" in self.data:
            if batch_send:
                return "email/batchWithTemplates"
            else:
                # This is the one Postmark API documented to have a trailing slash.
                # (Typo?)
                return "email/withTemplate/"
        else:
            # Postmark ignores TemplateModel unless using a template API.
            if "TemplateModel" in self.data:
                self.unsupported_feature("merge_global_data without template_id")
            if self.merge_data:
                self.unsupported_feature("merge_data without template_id")
            if batch_send:
                return "email/batch"
            else:
                return "email"
```

medmunds (changes requested):
@Sanjays2402 welcome, and thanks for this PR.

I think we also need to treat `merge_global_data` without `template_id` as unsupported. (That also uses Postmark's `TemplateModel` param.)

anymail/backends/postmark.py
Sanjays2402: Moved both checks into `get_api_endpoint()` as you laid out, pushed in 53daa6a. `TemplateModel` no longer selects a template endpoint on its own, so the non-template branch catches `merge_global_data` (either endpoint) and `merge_data`, and `serialize_data` is back to being just serialization.

Added `test_merge_global_data_without_template_unsupported` next to the merge_data one. Reverting the backend change makes both fail, so they are doing real work. `tests.test_postmark_backend` is 61 passing on Django 5.2.

medmunds (approved):
@Sanjays2402 looks good. Thanks again.

PR #480, opened 2026-07-29
Add MailKite ESP backend (transactional send)

bucabay:
This adds support for **MailKite** (https://mailkite.dev) as an Anymail ESP, following `ADDING_ESPS.md`. I'm opening this on behalf of MailKite, and happy to maintain the backend going forward.

MailKite is a transactional ESP with a straightforward JSON-over-HTTPS send API (`POST /v1/send`, Bearer auth). The backend is modeled on the existing **Resend** backend, which is the closest modern JSON/REST analog in the tree.

## What's included

**Backend** — `anymail/backends/mailkite.py`
`AnymailRequestsBackend` + `RequestsPayload`. Supported:
- `from` / `to` / `cc` / `bcc`, `subject`, text + html bodies
- `reply_to` (MailKite's `replyTo` is a single string — multiple addresses are comma-joined into one `Reply-To`)
- `extra_headers` (raw MIME header object), `attachments` (filename is required by the API; generated from mimetype if absent)
- `send_at` → `scheduledAt` (MailKite supports "send later")
- `template_id` + `merge_global_data` → `templateId` / `templateData` (server-rendered templates)
- `track_opens`, and `esp_extra` passed straight into the JSON body

Unsupported — surfaced as `AnymailUnsupportedFeature` rather than mis-handled:
- **inline attachments** — MailKite's send API has no content-id / inline field on attachments, so `cid:`-referenced images can't be expressed without silently breaking the reference
- **tags** and **metadata** — the API has no dedicated fields (custom headers remain available via `extra_headers`/`esp_extra`)
- **`merge_data` / `merge_metadata` / `merge_headers`** — single send endpoint, no per-recipient batch (empty `merge_data` is allowed as a no-op, consistent with the other backends)

Recipient status is parsed from MailKite's `{id, status}` response (`sent` → sent, `scheduled` → queued).

## Tests
`tests/test_mailkite_backend.py` — 22 mock tests covering standard Django email features, the Anymail-specific features, recipient-status parsing (both `sent` and `scheduled` responses), configuration errors, and each unsupported feature. All pass under Django 6.0 / Python 3.13:

```
$ python runtests.py tests.test_mailkite_backend
Ran 22 tests in 0.049s
OK
```

## Docs
`docs/esps/mailkite.rst` (install, settings, quirks, `esp_extra`, templates), added to the ESP toctree and the `esp-feature-matrix.csv`.

## Boilerplate
`pyproject.toml` (description, keywords, `[mailkite]` extra — no extra deps), `tox.ini` (test factor), per `ADDING_ESPS.md`.

## Deliberately out of scope (planned follow-ups)
- **Tracking and inbound webhooks.** MailKite itself supports both delivery tracking and inbound email (received messages delivered as signed webhooks); wiring those into Anymail's tracking/inbound signals is the natural next PR. The feature-matrix reflects that (tracking/inbound rows = "No" for now).
- **Live integration tests.** Would need a CI testing account / sandbox; happy to set that up if you'd like to move the support-status from "Unsupported" to "Full".

Happy to adjust anything to match your conventions. Thanks for Anymail — the extension points made this very pleasant to implement.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -32,6 +32,7 @@ Anymail currently supports these ESPs:
 * **MailerSend**
 * **Mailgun** (Sinch transactional email)
 * **Mailjet** (Sinch transactional email)
+* **MailKite** (inbound-first email platform)
 * **Mailtrap**
 * **Mandrill** (MailChimp transactional email)
 * **Postal** (self-hosted ESP)
--- a/anymail/backends/mailkite.py
+++ b/anymail/backends/mailkite.py
@@ -1,0 +1,215 @@
+import mimetypes
+
+from ..exceptions import AnymailRequestsAPIError
+from ..message import AnymailRecipientStatus
+from ..utils import (
+    BASIC_NUMERIC_TYPES,
+    CaseInsensitiveCasePreservingDict,
+    get_anymail_setting,
+)
+from .base_requests import AnymailRequestsBackend, RequestsPayload
+
+
+class EmailBackend(AnymailRequestsBackend):
+    """
+    MailKite (mailkite.dev) API Email Backend
+
+    MailKite is an inbound-first email platform: receive email as a webhook,
+    send with one API. This backend implements transactional *sending* through
+    MailKite's REST API. (Tracking and inbound webhook handling are not yet
+    implemented; see the PR notes.)
+    """
+
+    esp_name = "MailKite"
+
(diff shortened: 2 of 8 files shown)


bucabay:
Closed in favor of https://github.com/anymail/django-anymail/pull/481

PR #481, opened 2026-07-30
Add MailKite email backend (transactional send)

bucabay:
## What this adds

A new Anymail email backend for **MailKite** (`mailkite.dev`) — an inbound-first developer email platform (transactional send *and* parsed inbound over signed webhooks). This PR implements the **transactional send API** (`POST /v1/send`), modeled closely on the existing Resend and Postmark JSON backends.

It follows `ADDING_ESPS.md` end-to-end: backend + payload, mock tests, docs + feature matrix, and all the boilerplate entries (pyproject, README, tox, CI matrix).

## How to configure

```python
# settings.py
EMAIL_BACKEND = "anymail.backends.mailkite.EmailBackend"

ANYMAIL = {
    "MAILKITE_API_KEY": "<your mk_live_… key>",  # or MAILKITE_API_KEY at root
}
```

The `from` address must be on a domain whose ownership is verified in MailKite.

## What's supported

| Anymail feature | MailKite |
| --- | --- |
| from / to / cc / bcc / subject / text / html | ✅ |
| `reply_to` | ✅ (single string — multiple addresses are joined) |
| `extra_headers` | ✅ (MailKite `headers`, string values) |
| `metadata` / `tags` | ✅ (carried as JSON in `X-Metadata` / `X-Tags` headers — MailKite has no dedicated fields) |
| `attachments` | ✅ (base64 `content` + camelCase `contentType`) |
| `template_id` + `merge_global_data` | ✅ (→ `templateId` / `templateData`, server-rendered) |
| `esp_extra` | ✅ |
| `anymail_status` | ✅ (from `{id, status}`) |

Raised as `AnymailUnsupportedFeature` (the send API genuinely can't express these):

- `envelope_sender`, `send_at` (no scheduling field on `/v1/send`)
- `merge_data` / `merge_metadata` / `merge_headers` — MailKite sends **one** message to all recipients (there's no per-recipient batch), so per-recipient merge can't be expressed. Global `template_id` + `merge_global_data` is the supported path.
- inline attachments (the send API has no Content-ID field)
- `track_clicks` / `track_opens` (no per-message toggle; configure at account/domain level)

## How I tested

- **43 mock tests** (`tests/test_mailkite_backend.py`) — standard email features, Anymail features, every unsupported feature, API error handling, recipient-status parsing (incl. response `status` pass-through), session sharing, and missing-config errors. All pass.
- **Full suite green** — 1308 tests, 0 failures, no regressions from the config/matrix changes (80 live-integration tests skip without creds, as expected).
- **`black` / `flake8` / `isort` clean** on the new files (matching the project's pre-commit config).
- **Payload verified against the published contract** — the JSON this backend emits for a full-options message matches MailKite's documented `send-request` schema field-for-field (correct endpoint, Bearer auth, `to`/`cc` arrays, single-string `replyTo`, string-valued `headers`, base64 attachment `content` with camelCase `contentType`).

`tests/test_mailkite_integration.py` adds self-skipping live tests gated on `ANYMAIL_TEST_MAILKITE_API_KEY` / `ANYMAIL_TEST_MAILKITE_DOMAIN`, and the CI matrix + secret wiring are in place — you'd just add the secrets when you have a test account.

## Scope & roadmap

This is **send-only**, intentionally. MailKite's headline capability is *inbound* (parsed body + auth verdict as one signed webhook) and delivery-status tracking — adding Anymail inbound (`AnymailInboundEvent`) and tracking (`AnymailTrackingEvent`) webhook handling is the natural follow-up. I've marked both "Not yet" in the feature matrix and noted it in `docs/esps/mailkite.rst`. Happy to tackle inbound next if there's interest.

MailKite has a free tier, so live integration testing is straightforward.

## Maintenance

I'm with the MailKite team and can maintain this backend going forward — keeping it current with MailKite API changes, adding inbound/tracking support, and responding to issues.

---

*This contribution was prepared by MailKite's development team (with AI assistance, as disclosed in the commit trailer). No promotional or referral content is included in the code or docs.*

### DIFF
--- a/anymail/backends/mailkite.py
+++ b/anymail/backends/mailkite.py
@@ -198,9 +198,13 @@ class MailKitePayload(RequestsPayload):
             pass
         self.data["scheduledAt"] = send_at
 
-    # MailKite's /v1/send endpoint has no per-message tracking toggle.
+    def set_track_opens(self, track_opens):
+        # Per-send open-tracking override (HTML messages only). When omitted,
+        # the from-domain's default applies.
+        self.data["trackOpens"] = track_opens
+
+    # MailKite's /v1/send endpoint has no per-message click-tracking toggle.
     # def set_track_clicks(self, track_clicks):
-    # def set_track_opens(self, track_opens):
     # def set_envelope_sender(self, envelope_sender):
 
     def set_template_id(self, template_id):
--- a/anymail/backends/mailkite.py
+++ b/anymail/backends/mailkite.py
@@ -273,8 +273,12 @@ class MailKitePayload(RequestsPayload):
         # the from-domain's default applies.
         self.data["trackOpens"] = track_opens
 
-    # MailKite's /v1/send endpoint has no per-message click-tracking toggle.
-    # def set_track_clicks(self, track_clicks):
+    def set_track_clicks(self, track_clicks):
+        # Per-send click-tracking override (HTML messages only): MailKite
+        # rewrites links to a signed redirect that records the click. When
+        # omitted, the from-domain's default applies.
+        self.data["trackClicks"] = track_clicks
+
     # def set_envelope_sender(self, envelope_sender):
 
     def set_template_id(self, template_id):
--- a/anymail/backends/mailkite.py
+++ b/anymail/backends/mailkite.py
@@ -53,6 +53,34 @@ class EmailBackend(AnymailRequestsBackend):
 
     def parse_recipient_status(self, response, payload, message):
         parsed_response = self.deserialize_json_response(response, payload, message)
+
+        if isinstance(parsed_response, dict) and "results" in parsed_response:
+            # Batch send: per-recipient outcomes, in the same order as the
+            # recipients we posted. A batch can partially succeed -- failed
+            # recipients get status "failed" (no exception is raised).
+            results = parsed_response["results"]
+            if not isinstance(results, list) or len(results) != len(
+                payload.to_recipients
+            ):
+                raise AnymailRequestsAPIError(
+                    "Invalid MailKite API batch response format",
+                    email_message=message,
+                    payload=payload,
+                    response=response,
+                    backend=self,
+                )
+            recipient_status = CaseInsensitiveCasePreservingDict()
+            for recip, result in zip(payload.to_recipients, results):
+                status = result.get("status", DEFAULT_RESPONSE_STATUS)
+                if status == "failed" and result.get("code") == "recipient_suppressed":
+                    # unsubscribed / bounced / complained address
+                    status = "rejected"
+                status = RESPONSE_STATUS_MAP.get(status, status)
+                recipient_status[recip.addr_spec] = AnymailRecipientStatus(
+                    message_id=result.get("id"), status=status
+                )
+            return dict(recipient_status)
+
         try:
             message_id = parsed_response["id"]
             status = parsed_response.get("status", DEFAULT_RESPONSE_STATUS)
@@ -81,6 +109,11 @@ class EmailBackend(AnymailRequestsBackend):
 class MailKitePayload(RequestsPayload):
     def __init__(self, message, defaults, backend, *args, **kwargs):
         self.recipients = []  # for parse_recipient_status
+        self.to_recipients = []  # `to` only, for batch fan-out
+        self.metadata = None  # needed to merge with merge_metadata in batch
+        self.merge_data = {}
+        self.merge_metadata = {}
+        self.merge_headers = {}
         headers = kwargs.pop("headers", {})
         headers["Authorization"] = "Bearer %s" % backend.api_key
         headers["Content-Type"] = "application/json"
@@ -88,11 +121,45 @@ class MailKitePayload(RequestsPayload):
         super().__init__(message, defaults, backend, headers=headers, *args, **kwargs)
 
     def get_api_endpoint(self):
-        # MailKite has a single send endpoint; there is no separate batch API.
+        if self.is_batch():
+            # One personalized message per `to` recipient.
+            return "v1/send/batch"
         return "v1/send"
 
     def serialize_data(self):
+        if self.is_batch():
+            self.restructure_data_for_batch()
         return self.serialize_json(self.data)
+
+    def restructure_data_for_batch(self):
+        """Convert the single-send payload to MailKite's batch-send shape"""
+        # Shared fields (from/subject/html/text/templateId/templateData/headers/
+        # replyTo/attachments/scheduledAt/...) stay as built; the `to` list
+        # becomes recipients[], each entry carrying its own templateData and
+        # headers merged over the shared ones (per-recipient wins).
+        if "cc" in self.data or "bcc" in self.data:
+            # Batch sends one message per `to` recipient; there is no cc/bcc.
+            self.unsupported_feature("cc or bcc with batch send")
+        self.data.pop("to", None)
+        recipients = []
+        for email in self.to_recipients:
+            recipient = {"to": email.format(idna_encode=self.backend.idna_encode)}
+            recipient_data = self.merge_data.get(email.addr_spec)
+            if recipient_data:
+                recipient["templateData"] = recipient_data
+            headers = {}
+            if email.addr_spec in self.merge_metadata:
+                # Anymail semantics: this recipient's metadata is `metadata`
+                # updated with their merge_metadata entry.
+                recipient_metadata = dict(self.metadata or {})
+                recipient_metadata.update(self.merge_metadata[email.addr_spec])
+                headers["X-Metadata"] = self.serialize_json(recipient_metadata)
+            if email.addr_spec in self.merge_headers:
+                headers.update(self.merge_headers[email.addr_spec])
+            if headers:
+                recipient["headers"] = headers
+            recipients.append(recipient)
+        self.data["recipients"] = recipients
 
     #
     # Payload construction
@@ -112,6 +179,8 @@ class MailKitePayload(RequestsPayload):
                 email.format(idna_encode=self.backend.idna_encode) for email in emails
             ]
             self.recipients += emails
+            if recipient_type == "to":
+                self.to_recipients = emails
 
     def set_subject(self, subject):
         self.data["subject"] = subject
@@ -181,6 +250,7 @@ class MailKitePayload(RequestsPayload):
         self.data.setdefault("headers", {})["X-Metadata"] = self.serialize_json(
             metadata
         )
+        self.metadata = metadata  # merged with merge_metadata in a batch send
 
     def set_tags(self, tags):
         # MailKite has no tags field; carry them as JSON in a custom header.
@@ -214,11 +284,20 @@ class MailKitePayload(RequestsPayload):
         # MailKite renders templates server-side from templateData.
         self.data["templateData"] = merge_global_data
 
+    # Setting any of merge_data / merge_metadata / merge_headers switches to
+    # MailKite's batch-send endpoint: one personalized message per `to`
+    # recipient, with per-recipient values merged over the shared ones.
+    # The payload is restructured in serialize_data (see
+    # restructure_data_for_batch), once every attribute has been processed.
+
     def set_merge_data(self, merge_data):
-        # MailKite sends a single message to all recipients (not one per
-        # recipient), so per-recipient merge data can't be expressed.
-        if any(recipient_data for recipient_data in merge_data.values()):
-            self.unsupported_feature("merge_data")
+        self.merge_data = merge_data
+
+    def set_merge_metadata(self, merge_metadata):
+        self.merge_metadata = merge_metadata
+
+    def set_merge_headers(self, merge_headers):
+        self.merge_headers = merge_headers
 
     def set_esp_extra(self, extra):
         self.data.update(extra)
--- a/anymail/backends/mailkite.py
+++ b/anymail/backends/mailkite.py
@@ -11,9 +11,12 @@ from .base_requests import AnymailRequestsBackend, RequestsPayload
 
 # MailKite message status values returned from the send endpoint, mapped to
 # Anymail's normalized recipient statuses. The API documents the send response
-# as ``{"id": "...", "status": "..."}`` (e.g. ``"queued"``); we pass the status
-# through as-is, defaulting to ``"queued"`` when the ESP omits it.
+# as ``{"id": "...", "status": "..."}`` (e.g. ``"queued"``, or ``"scheduled"``
+# for a send_at message parked for later delivery); we normalize ``"scheduled"``
+# to Anymail's ``"queued"`` and pass other statuses through as-is, defaulting
+# to ``"queued"`` when the ESP omits the status.
 DEFAULT_RESPONSE_STATUS = "queued"
+RESPONSE_STATUS_MAP = {"scheduled": "queued"}
 
 
 class EmailBackend(AnymailRequestsBackend):
@@ -53,6 +56,7 @@ class EmailBackend(AnymailRequestsBackend):
         try:
             message_id = parsed_response["id"]
             status = parsed_response.get("status", DEFAULT_RESPONSE_STATUS)
+            status = RESPONSE_STATUS_MAP.get(status, status)
         except (KeyError, TypeError) as err:
             raise AnymailRequestsAPIError(
                 "Invalid MailKite API response format",
@@ -182,9 +186,22 @@ class MailKitePayload(RequestsPayload):
         # MailKite has no tags field; carry them as JSON in a custom header.
         self.data.setdefault("headers", {})["X-Tags"] = self.serialize_json(tags)
 
-    # MailKite's /v1/send endpoint has no scheduling parameter (scheduled sends
-    # are a separate flow), and no per-message tracking toggle.
-    # def set_send_at(self, send_at):
+    def set_send_at(self, send_at):
+        # A future scheduledAt parks the message for MailKite's scheduler;
+        # an omitted or past value sends immediately. Note: MailKite rejects
+        # scheduled sends that include custom headers (which this backend also
+        # uses to carry metadata and tags), so send_at can't currently be
+        # combined with extra headers, metadata, or tags.
+        try:
+            send_at = send_at.isoformat(
+                timespec="milliseconds" if send_at.microsecond else "seconds"
+            )
+        except AttributeError:
+            # User is responsible for formatting their own string
+            pass
+        self.data["scheduledAt"] = send_at
+
+    # MailKite's /v1/send endpoint has no per-message tracking toggle.
     # def set_track_clicks(self, track_clicks):
     # def set_track_opens(self, track_opens):
     # def set_envelope_sender(self, envelope_sender):
--- a/anymail/backends/mailkite.py
+++ b/anymail/backends/mailkite.py
@@ -188,10 +188,7 @@ class MailKitePayload(RequestsPayload):
 
     def set_send_at(self, send_at):
         # A future scheduledAt parks the message for MailKite's scheduler;
-        # an omitted or past value sends immediately. Note: MailKite rejects
-        # scheduled sends that include custom headers (which this backend also
-        # uses to carry metadata and tags), so send_at can't currently be
-        # combined with extra headers, metadata, or tags.
+        # an omitted or past value sends immediately.
         try:
             send_at = send_at.isoformat(
                 timespec="milliseconds" if send_at.microsecond else "seconds"
--- a/anymail/backends/mailkite.py
+++ b/anymail/backends/mailkite.py
@@ -1,0 +1,206 @@
+import mimetypes
+
+from ..exceptions import AnymailRequestsAPIError
+from ..message import AnymailRecipientStatus
+from ..utils import (
+    BASIC_NUMERIC_TYPES,
+    CaseInsensitiveCasePreservingDict,
+    get_anymail_setting,
+)
+from .base_requests import AnymailRequestsBackend, RequestsPayload
+
+# MailKite message status values returned from the send endpoint, mapped to
+# Anymail's normalized recipient statuses. The API documents the send response
+# as ``{"id": "...", "status": "..."}`` (e.g. ``"queued"``); we pass the status
+# through as-is, defaulting to ``"queued"`` when the ESP omits it.
+DEFAULT_RESPONSE_STATUS = "queued"
+
+
+class EmailBackend(AnymailRequestsBackend):
+    """
+    MailKite (mailkite.dev) API Email Backend
+
+    MailKite is an inbound-first email platform: it sends transactional mail over
+    a single JSON API and (unlike most ESPs) also *receives* mail — parsed body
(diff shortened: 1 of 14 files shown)


medmunds:
Thanks for the submission. People wanting support for MailKite in Anymail can use GitHub's 👍 reaction on the first post in this PR to express interest.

This seems to be a duplicate of PR #480 with some changes. Could you (or your agentic team) please close one of them?

I probably won't have a chance to review this closely until late August. In the meantime:
* If there are any updates, please just update this PR in place (rather than opening a new one).
* Please run the integration tests yourself, locally, using your own test account, and verify the email received. (It's helpful to temporarily change one of the *to* addresses in each integration test to your own email for this.)
* Your agents seem to have some disagreement over whether scheduled sending is supported (it was in the other PR, but not in this PR). If your product capabilities are still evolving rapidly, it might be a good idea to let it stabilize a bit before adding django-anymail integration. (As a reminder, so long as MailKite supports SMTP, it can be used with Django even without django-anymail integration.)

bucabay:
Scheduled sending (send_at) is now supported and pushed to this PR in place. Integration tests were run locally against a live MailKite account with receipt verified (including a send_at message delivered on schedule).

bucabay:
Thanks for the review notes @medmunds. Updates, all pushed to this PR in place:

Duplicate: #480 is closed — this PR is the only one going forward.

Scheduled sending is now supported. The earlier disagreement was real: MailKite's /v1/send accepts scheduledAt, but at the time it rejected scheduled sends that carried custom headers — and this backend transports metadata/tags as X-Metadata/X-Tags headers, so I'd left send_at unsupported rather than ship a half-working feature. We've since removed that limitation on the platform side (headers are now persisted and replayed at fire time), so the backend now implements set_send_at (ISO 8601 scheduledAt, same serialization approach as the Resend backend) with no caveats. The API's "scheduled" response status is normalized to Anymail's "queued". Docs and the feature matrix are updated to match.


Integration tests: run locally against the live API with our production account, including with a to address temporarily pointed at my own inbox as you suggested. All pass. I verified receipt of the actual emails end-to-end: the immediate sends, and a send_at (+2 min) message that was delivered on schedule with subject, body, and the metadata/tags headers intact (SPF/DKIM pass on the receiving side). The scheduled-send integration test also exercises metadata + tags together with send_at to keep that combination covered.

On product stability: fair point. The send API surface this backend depends on (/v1/send, scheduledAt, templates, attachments) is the stable core and is contract-tested against our published JSON schemas, so I don't expect further churn here. Happy to hold further changes until your review — no rush on our end, late August is fine.

Below are integration tests, the screenshots in Gmail and raw email headers/body. 

Screenshot 2026-07-31 at 3 45 13 PM
Screenshot 2026-07-31 at 3 44 50 PM

[Anymail MailKite scheduled-send integration test.eml](https://github.com/user-attachments/files/30588326/Anymail.MailKite.scheduled-send.integration.test.eml)

[Anymail MailKite integration test.eml](https://github.com/user-attachments/files/30588327/Anymail.MailKite.integration.test.eml)

bucabay:
Thanks for the earlier guidance @medmunds — a status update on everything pushed to this PR in place since (no new PRs; #480 stays closed).

**Feature support has filled out considerably.** Since your review note, MailKite's send API gained the pieces that were missing, and the backend now maps them:

- **Scheduled sending** (`send_at` → `scheduledAt`, ISO 8601, same serialization approach as the Resend backend). The API's `"scheduled"` status normalizes to `"queued"`. No caveats — custom headers (which this backend uses for metadata/tags) are carried on scheduled sends.
- **Inbound webhook** (`MailKiteInboundWebhookView` at `anymail/mailkite/inbound/`): HMAC-SHA256 signature verification (`X-MailKite-Signature`, timestamp tolerance, new `MAILKITE_WEBHOOK_SECRET` setting), mapping MailKite's pre-parsed JSON (no raw MIME) onto `AnymailInboundMessage` — envelope, display names, bodies, attachments (signed-URL and inlined-base64 variants), spam verdict → `spam_detected`, SPF/DKIM/DMARC via `esp_event`. Unknown event types are ignored for forward compatibility.
- **`track_opens`** as a per-message override of the domain default (click tracking remains unsupported and documented as such).
- **Batch sending**: `merge_data`, `merge_metadata`, and `merge_headers` now switch to MailKite's batch endpoint (`POST /v1/send/batch`) — one personalized message per `to` recipient, per-recipient message ids, partial-success reporting (`rejected` for suppressed addresses), max 50 recipients per batch, cc/bcc not combinable with batch. `merge_global_data` stays the shared default; per-recipient values win key-by-key.

**Testing.** Full suite passes locally (1337 tests, Python 3.12 / Django 6.0); lint clean. Integration tests run against the live API with our production account — including, per your suggestion, with a `to` address temporarily pointed at my own inbox — and I verified actual receipt end-to-end for each path: immediate sends, a `send_at` (+2 min) message delivered on schedule with headers intact, and a live batch received as two individual messages with correct per-recipient template rendering (SPF/DKIM pass on the receiving side). The inbound webhook tests use the exact payload shape MailKite's published `email.received` JSON schema and webhook-test endpoint produce, and the batch shape is contract-tested against our published schemas across nine SDK languages, so the surface this backend depends on is stable.

Remaining "No"s in the matrix (tracking-event webhooks, `track_clicks`, `envelope_sender`, AMP) reflect genuine ESP scope and are documented per row. Happy to adjust anything for the review — late August is fine on our end.

bucabay:
Two more features landed on the MailKite platform this week, and the backend now maps both — pushed to this PR in place:

**Click tracking (`track_clicks`).** MailKite added opt-in click tracking (links rewritten to a signed redirect that records the click and 302s to the original URL; `mailto:`/`tel:`/anchors never rewritten). The backend maps `track_clicks` to the per-send `trackClicks` override, exactly like `track_opens`. Verified live: a tracked send's links were rewritten in the received message, the redirect resolved to the exact original URL (query string intact), and the click was recorded.

**Status tracking webhooks (`AnymailTrackingEvent`).** MailKite now emits signed engagement events — `email.sent`, `email.bounced`, `email.complained`, `email.opened`, `email.clicked` — to an opt-in per-domain tracking webhook (documented at https://mailkite.dev/api/webhooks). New `MailKiteTrackingWebhookView` at `anymail/mailkite/tracking/`:

- Same signature scheme and account secret as the inbound webhook, so the one `MAILKITE_WEBHOOK_SECRET` setting covers both views (shared validation refactored into a base class).
- Bounces map to `reject_reason: bounced` with the DSN diagnostic in `mta_response`; complaints to `spam`; clicks carry `click_url`; opens/clicks carry `user_agent`. Unknown future event types normalize to `unknown` rather than erroring, and `email.delivered` (reserved by MailKite) is mapped ahead of time.
- Documented caveat: bounce/complaint events originate from provider notifications and carry a null `message_id` — `recipient` is the reliable key for those. MailKite's scanner/proxy machine-flags on opens and clicks ride along in `esp_event`.
- Swapped inbound/tracking URLs raise a helpful `AnymailConfigurationError` in both directions.

Suite is green (1353 tests), lint clean, live integration tests still pass against production. The feature matrix for MailKite is now Yes everywhere except `envelope_sender` and AMP — both genuine ESP scope, documented on the ESP page. The tracking-event payloads are contract-tested against MailKite's published JSON schema (https://api.mailkite.dev/v1/schemas/tracking-event.json), so the shape this webhook view depends on is pinned.

Nothing further planned before your review — the PR should be stable from here.

bucabay:
Hi @medmunds — gentle check-in, now that we're in late August. This PR is kept up to date in place: scheduled sending (`send_at`) is supported, the integration tests were run locally against a live account with receipt verified, and #480 stays closed so this is the only PR going forward. No rush — happy to make any changes whenever you've had a chance to look.

medmunds:
Hi @bucabay. I'm slowly working through backlog from before my holiday. This is on the list, but there are a bunch of other things before it.

anymail/backends/mailkite.py
```diff
@@ -0,0 +1,318 @@
...
+        filename = attachment.name or ""
+        if not filename:
+            # MailKite requires a filename. Generate a reasonable default.
+            ext = mimetypes.guess_extension(attachment.mimetype or "")
+            if ext:
+                filename = "attachment%s" % ext
+            else:
+                self.unsupported_feature(
+                    "unnamed attachments of type %s" % attachment.mimetype
+                )
```
medmunds, line 236: Please use f-strings. (Our minimum supported Python version is currently 3.10.)

anymail/backends/mailkite.py
```diff
@@ -0,0 +1,318 @@
...
+                self.make_attachment(attachment) for attachment in attachments
+            ]
+
+    def set_metadata(self, metadata):
+        # MailKite has no dedicated metadata field; carry it as JSON in a custom
+        # header (the ESP accepts arbitrary string-valued raw MIME headers).
+        self.data.setdefault("headers", {})["X-Metadata"] = self.serialize_json(
+            metadata
+        )
+        self.metadata = metadata  # merged with merge_metadata in a batch send
```
medmunds: Metadata is meant to be available to the app after sending (ideally in tracking webhooks, but at minimum somehow retrievable in the API or dashboard). From my testing:
* It seems like header values are not provided to webhooks or available in the API, so there is no point in constructing a custom header with them.
* MailKite's `/v1/send` API *does* have a documented `metadata` field, and those values can be accessed through the `GET /api/messages/:messageId` (though are not included directly in the webhook payload).
* The `/v1/send/batch` API does *not* support `metadata`.

Recommendation:
* Use the `metadata` API field for non-batch sends. Document that it can be retrieved if desired by calling the get message API from the tracking webhook handler. (I don't think we should try to automatically retrieve metadata in the MailKiteTrackingWebhookView, both to avoid the extra request and because it requires an account-level API key.)
* Treat `metadata` as an unsupported feature for batch sends (in `restructure_data_for_batch()`).

anymail/backends/mailkite.py
```diff
@@ -0,0 +1,318 @@
...
+
+    def set_tags(self, tags):
+        # MailKite has no tags field; carry them as JSON in a custom header.
+        self.data.setdefault("headers", {})["X-Tags"] = self.serialize_json(tags)
```
medmunds: The purpose of tags is to allow segmenting ESP analytics, or at minimum to be available in tracking webhook calls. A custom header doesn't support either of those.

Recommend treating tags as unsupported.

anymail/backends/mailkite.py
```diff
@@ -0,0 +1,318 @@
...
+    # MailKite's batch-send endpoint: one personalized message per `to`
+    # recipient, with per-recipient values merged over the shared ones.
+    # The payload is restructured in serialize_data (see
+    # restructure_data_for_batch), once every attribute has been processed.
+
+    def set_merge_data(self, merge_data):
+        self.merge_data = merge_data
+
+    def set_merge_metadata(self, merge_metadata):
+        self.merge_metadata = merge_metadata
```
medmunds: MailKite's current APIs cannot support `merge_metadata`. (See discussion in `metadata` above.)

anymail/webhooks/mailkite.py
```diff
@@ -0,0 +1,255 @@
...
+            if str(esp_event.get("type", "")).startswith("email."):
+                raise AnymailConfigurationError(
+                    "You seem to have set MailKite's *tracking-event* webhook"
+                    " to Anymail's MailKite *inbound* webhook URL."
+                    " (Or MailKite has added an event type this version of"
+                    " Anymail doesn't know about.)"
+                )
+            # Ignore anything else, rather than erroring on every delivery.
+            return []
+        return [self.esp_to_anymail_event(esp_event)]
```
medmunds: Since MailKite only supports a single webhook per domain, I think we'll need to use a combined webhook view that handles both inbound and tracking events. (See the `MandrillCombinedWebhookView` for an example. The MailKite version can be a little simpler since events are not batched and there are no compatibility concerns with earlier Anymail releases.)

docs/esps/mailkite.rst
```diff
@@ -0,0 +1,234 @@
...
+      ANYMAIL = {
+          ...
+          "MAILKITE_API_KEY": "<your MailKite API key>",
+      }
+
+Anymail will also look for ``MAILKITE_API_KEY`` at the root of the settings file
+if neither ``ANYMAIL["MAILKITE_API_KEY"]`` nor ``ANYMAIL_MAILKITE_API_KEY`` is set.
+
+You can override the API key for an individual message in its
+:ref:`esp_extra <mailkite-esp-extra>`.
```
medmunds: I don't think this is true—I didn't see any code for it. (And it seems unnecessary to support.)

docs/esps/mailkite.rst
```diff
@@ -0,0 +1,234 @@
...
+in your settings.py.
+
+
+.. setting:: ANYMAIL_MAILKITE_API_KEY
+
+.. rubric:: MAILKITE_API_KEY
+
+Required. A MailKite API key (an ``mk_live_…`` token), which you can create in the
+`MailKite dashboard`_. The sender address (``from``) must be on a domain whose
+ownership you've verified in MailKite.
```
medmunds, line 48: Do we want to recommend a domain-scoped API key here when possible? (Rather than the account-level one?)

docs/esps/mailkite.rst
```diff
@@ -0,0 +1,234 @@
...
+  and a batch can partially succeed — a suppressed address reports status
+  ``rejected`` and other failures ``failed``, without raising an error.
+  Two restrictions: MailKite allows at most **50 recipients per batch
+  message**, and ``cc``/``bcc`` can't be combined with a batch send (each
+  message goes to exactly one recipient).
+
+**Single reply-to field**
+  MailKite's ``replyTo`` is a single string. If you supply multiple reply-to
+  addresses, Anymail joins them into that one string (a header can hold several
+  addresses).
```
medmunds: This is an implementation detail that's unnecessary to mention. If the result is that multiple reply-to addresses work, it's not a limitation or quirk.

tests/test_mailkite_webhooks.py
```diff
@@ -0,0 +1,181 @@
+import json
```
medmunds, line 1: Please rename this file to `test_mailkite_webhooks.py` to parallel other ESPs. (I realize "tracking" would be more logical, but consistency simplifies maintenance.)

medmunds (changes requested):
Thanks @bucabay, I appreciate your patience. I've set up an Anymail test account on MailKite and explored the APIs a bit.

I pushed a couple of commits to this PR:
* Merged in current main branch
* Added a workaround for non-ASCII custom header values (and a test for non-ASCII headers in general)

And I've left review comments on some other things that need attention.

Incidentally, while testing international email support, I noticed a bug with Unicode bodies: MailKite seems to send 8bit utf-8 content (and correctly identifies the charset in the *Content-Type*), but incorrectly sets the CTE to 7bit. This may cause messages to be rejected or corrupted depending on the MTAs along the delivery path:

```eml
--mailkite-alt-kddxyvcz
Content-Type: text/plain; charset="UTF-8"
Content-Transfer-Encoding: 7bit

Prostý text
```

I'm guessing that's something you'll want to fix in MailKite's API, but if not we should probably note it in the "limitations and quirks" section of Anymail's docs.

Finally, send API calls seem to take several seconds (4000-8000+ ms). Is that expected?

bucabay:
Thanks for the thorough review — and for setting up an account and digging into the APIs. That found two real bugs on our side, not just doc issues. All review comments are addressed in the two commits just pushed.

## The Unicode CTE bug — you were right, and it was worse than reported

Our SES raw-MIME builder hardcoded the literal `'Content-Transfer-Encoding: 7bit'` on all six of its body-part branches while emitting the body string verbatim. A part may claim 7bit only when every octet is < 128 *and* no line exceeds 998 octets (RFC 2045 §6.2, RFC 5322 §2.1.1). Our self-hosted-relay builder had the same bug implicitly, by omitting the header — which RFC 2045 §6.1 defaults to 7bit.

Fixed by choosing the CTE the content actually qualifies for and encoding to match. We went with quoted-printable rather than `8bit`: it is 7-bit clean so no hop needs 8BITMIME, and it folds over-long lines as a side effect (a minified HTML body is one line of several thousand octets — already illegal regardless of charset, which we hadn't noticed either). Pure-ASCII bodies still go out as `7bit` with content untouched.

Two more of the same class turned up while fixing it:

* **Unencoded 8-bit header values** — the thing you worked around in 6c2ba95. Confirmed on the wire: SES silently re-encodes `Subject` for us, which is why only custom headers showed it. Now RFC 2047 encoded server-side, so your workaround is redundant (harmless, though — it only touches non-ASCII values, and an already-encoded ASCII value passes through untouched). **Non-ASCII attachment filenames** now use RFC 2231, per your note in `test_attachments`.
* **A header-injection hole.** A header value containing CR/LF went straight into the header block, so a caller-supplied header could append arbitrary headers — or, after a blank line, a body. Now sanitized.

These are committed with tests on our side. I'll follow up here once they're live in production so you can re-test — the fix isn't deployed as I write this, and I'd rather you not re-test against the old behaviour and conclude nothing changed.

## Send latency — no, 4–8s is not expected, and it's largely fixed

Thanks for flagging it; it turned out to be a real and measurable defect. Reproduced and measured against production (warm, repeated calls):

| Probe | Total | Adds |
|---|---|---|
| Unauthed 404 route | ~0.55s | Network + edge (TLS connect is 0.18s of it) |
| `GET /api/domains` | ~1.28s | + auth and one read — **two** DB statements |
| `POST /v1/send` with `scheduledAt` | ~2.3s | + the gate chain, no provider call |
| `POST /v1/send` immediate | 3.0–3.6s | + the synchronous SES `SendEmail` |

The useful number is the second row: two DB reads cost ~0.73s over the bare 404, so **one database round trip from this Worker costs 0.2–0.35s** — our D1 primary isn't in the colo serving the request. That makes the *number of serial statements* the dominant term, not the cost of any single query.

Looking at it with that lens, the send path was issuing **13 reads in 11 serial round trips** — and four of those were the same user row, four the same domain row, because each gate independently re-fetched them. Now fixed two ways: a per-request read memo, and overlapping the gates that have no dependency on each other (each still awaited at its original gate, so which error a blocked caller sees is unchanged).

| | Reads | Serial round trips |
|---|---|---|
| Before | 13 | **11** |
| After | 8 | **5** |

At the measured per-hop cost, six fewer serial hops should be worth well over a second. I'll post real numbers once it's deployed rather than quote the arithmetic.

Two things that remain: our Worker bundle is 16 MB, 12 MB of which is a docs corpus embedded for an internal agent — that's a cold-start cost and the likely explanation for the slow end of your 4000–8000ms range; it needs a separate change. And ~0.8–1.3s is SES's own `SendEmail`, which we await because the response carries a real provider message id.

## Review comments

* **f-strings** — done, including the `Bearer` header.
* **`metadata`** — now uses the `metadata` API field, verified round-tripping through `GET /api/messages/:messageId`. Documented that it can be retrieved from a tracking-webhook handler, with your note that this needs an account-level key. No auto-retrieval in the webhook view, per your recommendation. Treated as unsupported in `restructure_data_for_batch()`.
* **`tags`** — now unsupported. Agreed: a custom header gives neither analytics segmentation nor a webhook value, so it was pretending to support something it didn't.
* **`merge_metadata`** — now unsupported, same reason.
* **Combined webhook view** — added `MailKiteCombinedWebhookView` at `anymail/mailkite/`, and made it what the docs recommend. The event→`AnymailEvent` conversions moved into mixins shared with the split views.

  One correction: MailKite does support more than one webhook per domain. There's a separate per-domain tracking-webhook URL (the split layout, which is what the paired views were written for) *and* a single-URL mode where engagement events are opted into the domain's main webhook. So I've kept all three views — combined for the single-URL default, and the pair for the split layout — rather than dropping working configurations. Happy to delete the split pair if you'd still rather carry one view; it's a small change.
* **API key override in `esp_extra`** — you're right, that was never implemented. Claim removed.
* **Domain-scoped API key** — good call; now recommended, noting that an account-level key is needed only for account-wide APIs like retrieving a message.
* **Single reply-to** — removed. Multiple reply-to addresses work, so it isn't a quirk.
* **`test_mailkite_tracking.py`** → **`test_mailkite_webhooks.py`**.

Feature matrix updated (`tags` and `merge_metadata` → No, `metadata` → caveated). Integration tests pass against a live account, including the batch and scheduled-send cases. The 8 `test_amazon_ses_inbound` errors in a full local run are a missing optional dependency in my environment, unrelated to this branch.

bucabay:
Both issues you raised are fixed and deployed.

## Unicode CTE

Our SES raw-MIME builder hardcoded `Content-Transfer-Encoding: 7bit` on every body part while emitting the body verbatim. A part can only claim `7bit` when every octet is ASCII and no line exceeds 998 octets, so any 8-bit body was mislabelled.

Fixed: each part now carries the encoding it qualifies for — quoted-printable for 8-bit UTF-8, `7bit` only when the content genuinely is. Verified end-to-end against production; the body decodes cleanly, so label and content agree.

Two related bugs fell out of the same builder and are also fixed:

- **Custom header values weren't RFC 2047 encoded** — the thing you worked around in 6c2ba95. `X-Custom: Prostý` now goes out as `=?UTF-8?B?UHJvc3TDvQ==?=`. Your workaround is no longer needed, but it's harmless if you'd rather keep it — it only touches non-ASCII values, and an already-encoded value passes through untouched.
- **Non-ASCII attachment filenames** now use RFC 2231, per your note in `test_attachments`.

## Send latency

Not expected, and a real defect. The gate chain was making **11 serial database round trips**, four of them re-reading the same user row and four the same domain row. Now **5**, by memoizing those reads per request and overlapping the gates that don't depend on each other.

Profiled on `POST /v1/send` after the fix (n=10):

| | Before | After |
|---|---|---|
| p50 | 3.0–3.6s | **0.97s** |
| p95 | — | 1.06s |
| Gate chain only (scheduled send, no provider call) | 2.30s | **0.53s** |

Conditions were better when I re-measured — an endpoint I didn't touch also improved — so not all of that delta is the fix. Isolating by time above baseline, the gate chain went +1.75s → +0.42s. The condition-independent number is the round-trip count, 11 → 5, which now has a test budgeting it.

Most of the remaining second is the synchronous SES call, which we await because the response carries a real provider message id.

## Review comments

All addressed in the two commits pushed earlier: `metadata` moved to MailKite's API field (unsupported for batch), `tags` and `merge_metadata` now unsupported, combined webhook view added, docs corrections, f-strings, test file renamed. Integration tests pass against a live account.

Thanks for pushing on the MIME issues — Gmail accepts the mislabelled message without complaint, so nothing surfaced it until you read the raw source.

---

*Edit: made this comment more concise.*

bucabay:
Heads up: I force-pushed this branch (`1834b40` → `22fbfc3`).

Two reasons — `main` requires verified signatures and none of my commits were signed, and the branch had fallen 2 commits behind. So I rebased onto `787c7ed` with SSH signing enabled. All 11 commits now verify.

Your `6c2ba95` is preserved as `8cf5c77`, still authored by you. Your merge commit `37aae1c` is gone, since the rebase supersedes it and `main` is linear anyway.

No content changed — the diff over every MailKite file is identical to before the rebase. The only edit was resolving `anymail/urls.py`, where Mailtrap's inbound webhook had landed in the same spot as MailKite's; both are registered now. Checks are green.

PR #482, opened 2026-08-03
Add MailKite ESP backend

bucabay:
## What this adds

A new **MailKite** ([mailkite.dev](https://mailkite.dev)) backend, modeled closely on the existing **Resend** backend — MailKite's send API is JSON/shape-compatible with Resend's (`POST` JSON, `Authorization: Bearer`, `{from,to,subject,html,text,…}`), so the contribution follows the same structure you already accept and maintain.

MailKite is an *inbound-first* developer email platform: it sends transactional mail, and also **receives** email (parsed body + SPF/DKIM/DMARC verdict) delivered to a signed webhook. This PR implements the **send** side.

### Anymail features supported

MailKite's API actually exposes a few things Resend's doesn't, so the backend supports **more** Anymail features than the Resend one:

| Feature | MailKite | (Resend, for reference) |
|---|---|---|
| `template_id` + `merge_global_data` (server-side templates, `{{merge_tags}}`) | ✅ | ❌ |
| `track_opens` / `track_clicks` (per-message override) | ✅ | ❌ |
| `send_at` (`scheduledAt`) | ✅ | ✅ |
| `merge_metadata` / `merge_headers` (batch send, `POST /v1/send/batch`) | ✅ | ✅ |
| `metadata`, `tags` (via `X-Metadata` / `X-Tags` headers) | ✅ | ✅ |
| attachments (`content`/`filename`/`contentType`) | ✅ | ✅ |

### Documented limitations (on the ESP page, raised as `AnymailUnsupportedFeature`)

- **No outbound tracking webhooks** — MailKite does not currently emit delivery/open/click/bounce events as webhooks, so `AnymailTrackingEvent` signals are not available. (Open/click tracking still works in MailKite's own dashboard, and per-message `track_opens`/`track_clicks` are supported.)
- **No inbound via Anymail signals** — MailKite *does* inbound (its headline feature), but delivers to the user's own webhook URL in its own payload format, so it isn't wired through `AnymailInboundEvent`.
- **No inline attachments (Content-ID)** — MailKite's send API has no `cid` field, so inline images raise rather than silently degrading to regular attachments.
- **Single `replyTo` only** — MailKite takes one reply-to string; multiple `reply_to` addresses raise.
- No `envelope_sender`.

## How to configure

```python
# settings.py
EMAIL_BACKEND = "anymail.backends.mailkite.EmailBackend"

ANYMAIL = {
    "MAILKITE_API_KEY": "mk_live_...",   # or MAILKITE_API_KEY / ANYMAIL_MAILKITE_API_KEY
}
# Optional: ANYMAIL = { ..., "MAILKITE_API_URL": "https://api.mailkite.dev/" }
```

No extra install — MailKite uses only Anymail's existing `requests` dependency (hence `mailkite = []` in optional-dependencies, no `[mailkite]` extra).

## How I tested it

- **`tests/test_mailkite_backend.py`** — 45 mocked tests covering standard Django email features and all the Anymail additions above, including the batch path (empty `merge_data`, `merge_metadata`, `merge_headers`) and a **per-recipient rejection** (MailKite reports batch failures per-recipient; verified they surface as `"rejected"`).
- **`tests/test_mailkite_integration.py`** — live tests against the real API, guarded by `ANYMAIL_TEST_MAILKITE_API_KEY` / `ANYMAIL_TEST_MAILKITE_DOMAIN` (skip cleanly without them).
- Ran the full mocked suite locally (Django 6.0, Python 3.13): `python runtests.py tests.test_mailkite_backend tests.test_mailkite_integration` → **49 passed (4 skipped)**. The Resend backend tests still pass unchanged.
- `pre-commit` (black, isort, flake8, doc8, pygrep-hooks) is clean on every file in this PR.

I did **not** run the live integration tests (I don't want to charge a real MailKite account from a PR). If you'd like, I can run them against a MailKite test domain before merge, or you can run them in CI with credentials.

## Files

- `anymail/backends/mailkite.py`
- `tests/test_mailkite_backend.py`, `tests/test_mailkite_integration.py`
- `docs/esps/mailkite.rst` (+ toctree, README, feature-matrix column, `pyproject` keywords/extra, changelog)

## Maintenance

Happy to maintain this backend going forward — keeping it in step with MailKite's send API as it evolves, and adding outbound tracking webhook support (→ `AnymailTrackingEvent`) and Anymail inbound signal support if/when MailKite exposes the corresponding surfaces. I've read `CONTRIBUTING` and followed the existing ESP-backend pattern; happy to rework anything that doesn't match your conventions.

(Heads-up on the feature-matrix support-status row: I marked MailKite `Unsupported` per the footnote — Anymail doesn't have a MailKite account for live integration tests yet. Adjust to `Limited`/`Full` as you see fit once you have access.)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -25,6 +25,22 @@ Release history
 ^^^^^^^^^^^^^^^
     ..  This extra heading level keeps the ToC from becoming unmanageably long
 
+Unreleased
+----------
+
+New ESPs
+~~~~~~~~
+
+* **MailKite:** Add support for this ESP. MailKite is an inbound-first developer
+  email platform -- it sends transactional mail via a single JSON API, and also
+  *receives* email (parsed body plus an authentication verdict, delivered as a
+  signed webhook). The backend supports per-message open/click tracking
+  overrides, server-side templates with merge tags (``template_id`` and
+  ``merge_global_data``), scheduled sending (``send_at``), and batch sending with
+  per-recipient ``merge_metadata`` and ``merge_headers``. It does not wire up
+  outbound tracking webhooks (MailKite does not emit them) or inbound signals
+  (MailKite delivers inbound mail directly to your own webhook URL).
+
 v15.1
 -----
 
--- a/README.rst
+++ b/README.rst
@@ -32,6 +32,7 @@ Anymail currently supports these ESPs:
 * **MailerSend**
 * **Mailgun** (Sinch transactional email)
 * **Mailjet** (Sinch transactional email)
+* **MailKite** (inbound-first email platform)
 * **Mailtrap**
 * **Mandrill** (MailChimp transactional email)
 * **Postal** (self-hosted ESP)
--- a/anymail/backends/mailkite.py
+++ b/anymail/backends/mailkite.py
@@ -1,0 +1,303 @@
+import mimetypes
+
+from ..exceptions import AnymailRequestsAPIError
+from ..message import AnymailRecipientStatus
+from ..utils import (
+    BASIC_NUMERIC_TYPES,
+    CaseInsensitiveCasePreservingDict,
+    get_anymail_setting,
+)
+from .base_requests import AnymailRequestsBackend, RequestsPayload
+
+
+class EmailBackend(AnymailRequestsBackend):
+    """
+    MailKite (mailkite.dev) API Email Backend
+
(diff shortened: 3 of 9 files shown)


bucabay:
Apologies — this duplicates #481, which is the maintained MailKite backend PR and the only one going forward (as promised there). This one was opened in error by our tooling; we've fixed the process so it can't happen again. Closing.

PR #487, opened 2026-09-05
Brevo: Update documented HTML email warning

francoisfreitag:
The Brevo API now accepts payloads containing only the `textContent` parameter for the email body. However, it still misbehaves by silently converting the plain text to HTML.

Update the warning in the documentation to better describe to the current behavior.


francoisfreitag:
Note: I work with @deronnax (the author of #483), the findings on Brevo current behavior are shared.

medmunds:
@francoisfreitag and @deronnax thanks for the info! (A lot of ESPs will convert text to html or sometimes html to text if one is missing, but this is the first I've seen that removes the text content after the html conversion.)

I shortened the note a bit for Anymail's docs.

PR #493, opened 2026-09-16
SendGrid: don't include charset in attachment type

### LINKED ISSUE #492: SendGrid: text attachments rejected since v14.0 ("The attachment type cannot contain ';', or CRLF characters.")
## Summary

Since django-anymail 14.0, the SendGrid backend sends the *full* Content-Type header of an attachment (including `; charset="utf-8"`) as the `type` field of the `/v3/mail/send` request. SendGrid requires a bare `maintype/subtype` there and rejects the whole message. Because `anymail.utils.Attachment` always adds a charset for `text/*` attachments, **every text attachment** (`.ics`, `.txt`, `.csv`, …) now fails to send via SendGrid.

I'm aware SendGrid support is officially dropped (#432). Filing this anyway because the regression is a one-line change and v13.1 behaved correctly.

## Environment

- django-anymail 14.0 (bug is still present on `main` / v15.2, see below)
- Django 5.2.13
- Python 3.12
- requests 2.33.1

## Steps to reproduce

```python
from django.core.mail import EmailMessage, get_connection

msg = EmailMessage(
    "Subject", "Body", "from@example.com", ["to@example.com"],
    connection=get_connection("anymail.backends.sendgrid.EmailBackend"),
)
msg.attach("file.ics", "BEGIN:VCALENDAR\r\nEND:VCALENDAR\r\n", "text/calendar")
msg.send()
```

Payload inspection without hitting the API:

```python
from anymail.backends.sendgrid import EmailBackend, SendGridPayload

backend = EmailBackend(api_key="x")
payload = SendGridPayload(msg, {}, backend)
print(payload.data["attachments"][0]["type"])
```

| anymail version | `attachments[0]["type"]` |
|---|---|
| 13.1 | `text/calendar` |
| 14.0 | `text/calendar; charset="utf-8"` |

## Actual result

```
anymail.exceptions.AnymailRequestAPIError: Sending a message to to@example.com from from@example.com
SendGrid API response 400 (Bad Request):
{"errors":[{"message":"The attachment type cannot contain ';', or CRLF characters.","field":"attachments.0.type","help":null}]}
```

Binary attachments (e.g. `image/png`) are unaffected because `Attachment.content_type` has no charset parameter for them.

## Expected result

The message is sent and the attachment `type` is `text/calendar`, as in v13.1.

## Root cause

`SendGridPayload.add_attachment` changed between v13.1 and v14.0.

v13.1:

```python
att = {
    "content": attachment.b64content,
    "type": attachment.mimetype,
    ...
}
```

v14.0 and `main`:

```python
# SendGrid strips charset from `type` and doesn't include `charset`
# in the Content-Type header, so use utf-8 encoding for text and hope
# for the best. (See issue #150.)
content_type = attachment.content_type
if attachment.charset and attachment.charset != "utf-8":
    # Rebuild content_type with charset=utf-8 to match what we'll send
    temp = MIMEPart()
    temp.add_header("Content-Type", attachment.mimetype, charset="utf-8")
    content_type = temp["Content-Type"]

att = {
    "content": attachment.b64content_utf8,
    "type": content_type,
    ...
}
```

`Attachment.content_type` is the full header value ("the full content-type, with charset param if text", per the docstring) and `Attachment.__init__` always sets a charset when `maintype == "text"` or the content is a `str`. So `type` always ends up with `; charset=...` for text attachments. The comment in the new code even says SendGrid strips the charset from `type`; in practice SendGrid rejects it outright rather than stripping it.

This looks like a side effect of the v14.0 fix "Fix Anymail bugs that could cause text attachments with Unicode content to display incorrectly in some email clients", which was applied to several ESPs and appears to have been applied to SendGrid too without a live test (SendGrid is no longer tested).

## Suggested fix

Keep the UTF-8 content encoding from v14.0 but send the bare mimetype as `type`, since SendGrid can't carry a charset there anyway:

```python
def add_attachment(self, attachment):
    # SendGrid's `type` must be a bare maintype/subtype (no parameters), and
    # it doesn't include a `charset` in the Content-Type header, so use
    # utf-8 encoding for text and hope for the best. (See issue #150.)
    att = {
        "content": attachment.b64content_utf8,
        "type": attachment.mimetype,
        # (filename is required -- submit empty string if unknown)
        "filename": attachment.name or "",
    }
    if attachment.inline:
        att["disposition"] = "inline"
        att["content_id"] = attachment.cid
    self.data.setdefault("attachments", []).append(att)
```

The `charset != "utf-8"` branch becomes unnecessary because `type` no longer carries a charset.

## Workaround

Subclass the backend and override `add_attachment` with the code above:

```python
from anymail.backends.sendgrid import EmailBackend as AnymailSendGridBackend
from anymail.backends.sendgrid import SendGridPayload as AnymailSendGridPayload


class SendGridPayload(AnymailSendGridPayload):
    def add_attachment(self, attachment):
        att = {
            "content": attachment.b64content_utf8,
            "type": attachment.mimetype,
            "filename": attachment.name or "",
        }
        if attachment.inline:
            att["disposition"] = "inline"
            att["content_id"] = attachment.cid
        self.data.setdefault("attachments", []).append(att)


class EmailBackend(AnymailSendGridBackend):
    def build_message_payload(self, message, defaults):
        return SendGridPayload(message, defaults, self)
```

Happy to open a PR with the fix and a test asserting `";" not in att["type"]` if you'd accept one for an unsupported ESP.

Adman:
* Since v14.0, the SendGrid backend sent the full Content-Type (with charset) as the attachment `type`. SendGrid requires a bare mimetype there and rejects the message: "The attachment type cannot contain ';', or CRLF characters." Send the bare mimetype again; text attachments are still encoded as utf-8.
* fixes #492


medmunds:
I updated the wording in the docs a bit to reflect my uncertainty about different SendGrid customers seeing different versions of the API. (See my recent comment in #492.) And I think it's sufficient to include this in the existing attachments test case that you updated, without adding a separate case. (But I did copy in your "no semicolon" assertion to avoid potential future regressions.)

Thanks again for the report and the PR!

